PR #2, opened 2018-03-09
Fix overflow bug

jimpo:
This could be used to drain a contract of its collateral by issuing a large quantity at once to trigger an overflow, then redeeming smaller amounts.


justinkchen:
LGTM, thanks for this!

Issue #5, opened 2018-04-01
Support Sets with Decimal Place Component Quantities


felix2feng:
# **Motivation**:
We need to provide more flexibility in the token quantities that can be instantiated as they currently can only be done in whole quantities. For example, we are currently able to instantiate a Set with 1x Token A and 2x Token B with whole numbers:
```
Deploy with the following parameters([`address_token_a`, `address_token_b`],[1, 2], 'AB_Set`, `AB`)
```

If the price of Token A and Token B are sufficiently different, it will require large units of Token A and Token B to achieve certain percentage splits at a clean price. In the case that Token A is $100 and Token B is $1 and we want to achieve a 50/50 split, we'd need to have `1x` Token A and `100x` Token B. This results in a total Set price of $200. If a user wanted to purchase $1 of a Set, they'd need to purchase 0.005 of this Set, a figure that is psychologically unappealing. 

If we wanted to represent the Set at a price of $1, we need to support decimal place representations in quantities with Set instantiation. In the same example, if we wanted to represent `AB_Set` with the same mix and have an initial price of $1, we can instantiate the Set with `0.05` Token A when Token A is $100 and `0.5` Token B that is $1. 

# **Solution**
In solidity, the standard ERC20 token supports up to 18 decimal places. We can support the same number in a Set.

Note: We do need to be mindful of the individual components and their decimal place limitations. For example, TRON is a ERC20 token that only supports 6 decimal places.

PR #8, opened 2018-04-04
Support Sets with Decimal Place Component Quantities

felix2feng:
This PR addresses the following [issue](https://github.com/SetProtocol/set-protocol-contracts/issues/5).

Major Change:
- Units must now be denominated in units of gWei (10^9). This means that if you intend for a unit of 1, you'd have to declare it as 1 x 10^9.
- These changes affected the `issue` and `redeem` functions
- Updated unit tests to test and reflect these changes.

Other Changes:
- Imported SafeMath as a default for uint256
- Made error messages of unit assertions more descriptive


package.json
```diff
@@ -7,7 +7,7 @@
     "test": "test"
   },
   "scripts": {
-    "test": "truffle test"
+    "test": "rm -rf build && truffle test"
```
asoong: Is this standard?

PR #9, opened 2018-04-08
Alex/add typescript

asoong:
- Converted tests to Typescript
- Upgrade solidity version to 0.4.21
- Generate Typings from Contract ABIs
- Set up configs
- Added solidity and typescript linting


test/config/bignumber_setup.ts
```diff
@@ -0,0 +1,11 @@
+import { BigNumber } from "bignumber.js";
```
justinkchen: since we're already including `"@0xproject/utils": "^0.1.0",` can we just use their exported BigNumber? Or I guess we could use our own for more flexibility

test/setToken-base.spec.ts
```diff
@@ -0,0 +1,355 @@
...
+// Contract types
+// import { StandardTokenMockContract } from "../types/generated/standard_token_mock";
+// import { SetTokenContract } from "../types/generated/set_token";
+
+// Artifacts
+const SetToken = artifacts.require("SetToken");
+const StandardTokenMock = artifacts.require("StandardTokenMock");
+
+// Testing Set up
+import { BigNumberSetup } from "./config/bignumber_setup";
```
justinkchen: we can move this into some `jest` configuration: https://facebook.github.io/jest/docs/en/configuration.html#setupfiles-array

justinkchen: nvm this is using truffle/mocha...i wonder if this file structure would work to pull this out to a common place https://stackoverflow.com/questions/24153261/joining-tests-from-multiple-files-with-mocha-js/26791892#26791892

PR #15, opened 2018-04-10
Partial Redeem Implementation and Happy Path Tests

felix2feng:
SetToken Contract
- Avoided duplication by adding some modifiers
- Added new mappings to track excluded balances
- Wrote partial Redeem and redeemExcluded

Wrote unit test

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "deploy:development": "bash scripts/deploy_development.sh",
     "dist": "bash scripts/prepare_dist.sh",
     "clean": "rm -rf build",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "deploy:development": "bash scripts/deploy_development.sh",
     "dist": "bash scripts/prepare_dist.sh",
     "clean": "rm -rf build",
+    "compile": "truffle compile",
```
justinkchen: could just make the `truffle compile --all` and put it into the `test` method below too

PR #22, opened 2018-05-02
Clarify private method for calcualting transfer value

asoong:
`Current` is only relevant to the scope of the calling block. The private method is for calculating transfer value of a component for a given quantity.


felix2feng:
👍

PR #38, opened 2018-06-11
Add pre/post transfer balance checks

asoong:
- Pre/post transfer balance checks.

### DIFF
--- a/contracts/core/TransferProxy.sol
+++ b/contracts/core/TransferProxy.sol
@@ -20,6 +20,7 @@ pragma experimental "ABIEncoderV2";
 
 import { Authorizable } from "../lib/Authorizable.sol";
 import { ERC20 } from "zeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
 
 
 /**
@@ -33,6 +34,9 @@ import { ERC20 } from "zeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 contract TransferProxy is
     Authorizable
 {
+    // Use SafeMath library for all uint256 arithmetic
+    using SafeMath for uint256;
+
     /* ============ State Variables ============ */
 
     // Address of the Vault contract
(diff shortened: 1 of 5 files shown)


contracts/core/TransferProxy.sol
```diff
@@ -76,11 +80,18 @@ contract TransferProxy is
         external
         onlyAuthorized
     {
+        // Retrive current balance of token for the vault
```
felix2feng: "Retrieve"

contracts/core/TransferProxy.sol
```diff
@@ -76,11 +80,18 @@ contract TransferProxy is
...
+        uint exstingVaultBalance = ERC20(_tokenAddress).balanceOf(vaultAddress);
```
felix2feng: "existingVaultBalance" vs "exstingVaultBalance"?

felix2feng: This works for now. Though, did you have any thoughts re: using something like dYdX's TokenInteract?

contracts/core/Vault.sol
```diff
@@ -92,11 +92,18 @@ contract Vault is
         isNonZero(_quantity)
         isValidDestination(_to)
     {
+        // Retrive current balance of token for the vault
```
felix2feng: Same as above

contracts/core/Vault.sol
```diff
@@ -92,11 +92,18 @@ contract Vault is
...
+        uint exstingVaultBalance = ERC20(_tokenAddress).balanceOf(this);
```
felix2feng: Same as above

test/transferProxy.spec.ts
```diff
@@ -155,16 +178,19 @@ contract("TransferProxy", (accounts) => {
...
+      await approveTransfer(mockToken, transferProxy.address, approver);
     });
 
     // Subject
-    const amountToTransfer = STANDARD_INITIAL_TOKENS;
+    const amountToTransfer: BigNumber = STANDARD_INITIAL_TOKENS;
+    let tokenAddress: Address;
 
     async function subject(): Promise<string> {
+      const tokenToTransfer = tokenAddress || mockToken.address;
```
felix2feng: Maybe add a comment?

test/vault.spec.ts
```diff
@@ -113,6 +108,26 @@ contract("Vault", (accounts) => {
       );
   };
 
+  async function deployTokenWithFee(fee: BigNumber, initialAccount: Address, from: Address = ownerAccount) {
```
felix2feng: I wonder if there's a way to make this a library so there's no duplication across `vault.spec` and `transferProxy.spec`

asoong: For sure, we will DRY up testing shortly.

PR #39, opened 2018-06-12
Refactor set token, removing functionality that is moving to other co…

asoong:
Gutting functionality moving to other contracts.


contracts/core/SetToken.sol
```diff
@@ -1,110 +1,126 @@
...
  * @author Felix Feng
+ *
  * @dev Implementation of the basic {Set} token.
+ *
  */
 contract SetToken is
     StandardToken,
-    DetailedERC20("", "", 18),
-    ISetToken
+    DetailedERC20("", "", 18)
```
felix2feng: To be careful here, we may want to be passing in the details in the constructor of the name and symbol. This means we may need to deprecate usage of DetailedERC20

contracts/core/SetToken.sol
```diff
@@ -1,110 +1,126 @@
...
+    modifier isMultipleOfNaturalUnit(uint _quantity) {
+        require((_quantity % naturalUnit) == 0);
         _;
     }
 
-    modifier validDestination(address _to) {
-        require(_to != address(0));
-        require(_to != address(this));
+    // Confirm that all inputs for creating a set are valid
+    modifier areValidCreationParameters(address[] _components, uint[] _units) {
```
felix2feng: Perhaps we want to do these checks in the `Factory` vs. in the newly created `SetToken` contract

contracts/core/SetToken.sol
```diff
@@ -152,167 +168,30 @@ contract SetToken is
...
-    function redeem(uint _quantity)
-        public
-        isMultipleOfNaturalUnit(_quantity)
-        hasSufficientBalance(_quantity)
-        isNonZero(_quantity)
-        returns (bool)
+    function mint(
+        uint quantity
+    )
+        external
```
felix2feng: I'm sure you're already aware, but this should only be callable by `Core`

asoong: Yup, we haven't extended Authorizable yet.

PR #40, opened 2018-06-12
Brian/authorizable

bweick:


test/logs/Authorizable.ts
```diff
@@ -0,0 +1,35 @@
...
+    event: "AddressAuthorized",
+    address: contractAddress,
+    args: {
+      authAddress,
+      authorizedBy,
+    },
+  }];
+};
+
+export function getExpectedRmvAuthorizedLog(
```
asoong: Let's be explicit here, we're not saving anything by short handing `remove`

test/authorizable.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+        index = new BigNumber(3);
+      });
+
+      it("should revert", async () => {
+        await expectRevertError(subject());
+      });
+    });
+
+    describe("when the passed index does not match target address", async () => {
+      beforeEach(async () => {
```
asoong: before?

test/authorizable.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+    });
+
+    it("removes address from authorities array", async () => {
+      const oldAuthoritiesArray = await authorizable.getAuthorizedAddresses.callAsync();
+
+      await subject();
+
+      const newAuthoritiesArray = await authorizable.getAuthorizedAddresses.callAsync();
+
+      oldAuthoritiesArray.splice(index.toNumber(), 1);
```
asoong: Let's define some variables to clarify what's going on here

test/authorizable.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+      expect(newAuthoritiesArray).to.eql(oldAuthoritiesArray);
+      expect(newAuthoritiesArray.length).to.eql(oldAuthoritiesArray.length);
```
asoong: This is unnecessary, it will be implicit in the equality of the previous expectation

test/authorizable.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+      await subject();
+
+      const storedAuthAddress = await authorizable.authorized.callAsync(
+        authorizedAccount,
+      );
+
+      expect(storedAuthAddress).to.eql(false);
+    });
+
+    it("removes address from authorities array", async () => {
```
asoong: This test case can be simplified with some nifty getters: http://www.chaijs.com/api/bdd/

test/authorizable.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+      const newAuthoritiesArray = await authorizable.getAuthorizedAddresses.callAsync();
+
+      oldAuthoritiesArray.splice(index.toNumber(), 1);
+      expect(newAuthoritiesArray).to.eql(oldAuthoritiesArray);
+      expect(newAuthoritiesArray.length).to.eql(oldAuthoritiesArray.length);
+    });
+
+    it("emits correct AuthorizedAddressRemoved log", async () => {
+      const txHash = await subject();
+      const formattedLogs = await getFormattedLogsFromTxHash(txHash);
```
asoong: Let's see if we can standardize log testing. It would be nice to have a helper function that takes a transaction hash, and the three parameters to `getExpectedRmvAuthorizedLog` and does the expectation. See `tokenAssertions.ts`

PR #41, opened 2018-06-12
Create CoreWrapper abstraction for DRYing up test setup

asoong:
Abstracting out some of the test setup into a CoreWrapper class to clean up some of the spec files. In the future there should be logical groupings/other wrappers (ERC20, Authorizable methods for example)

Currently only applied to `transferProxy.spec.ts`, let me know your thoughts and we can apply to other spec files.

### DIFF
--- a/test/constants/constants.ts
+++ b/test/constants/constants.ts
@@ -1,9 +1,10 @@
 import { BigNumber } from "bignumber.js";
 import { ether, gWei } from "../utils/units";
 
+export const DEFAULT_GAS = 7000000;
 export const INVALID_OPCODE = "invalid opcode";
-export const REVERT_ERROR = "revert";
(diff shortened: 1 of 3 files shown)


test/utils/coreUtils.ts
```diff
@@ -0,0 +1,131 @@
...
+      { from },
+    );
+  }
+
+  public async deployTransferProxyAsync(
+    vaultAddress: Address,
+    from: Address = this._tokenOwnerAddress
+  ): Promise<TransferProxyContract> {
+    const truffleTransferProxy = await TransferProxy.new(
+      { from, gas: 7000000 },
```
justinkchen: lets start putting this `gas` amount into a constant like `DEFAULT_GAS` or something

PR #43, opened 2018-06-12
Fixed keccak, exponential type casting, and other small warnings.

bweick:
Main warnings of concern were an [implicit type casting](https://github.com/ethereum/solidity/issues/1738) of exponentials and a deprecation warning for [packing variables for hashing](https://github.com/ethereum/solidity/issues/3955) that felt it was a good idea to just fix now since it seems it will be changed in the next solidity update.


asoong:
LGTM, can you also go ahead and remove the ABI Encoder messages?

PR #44, opened 2018-06-12
Add Factory with Create method

felix2feng:
- Wrote SetTokenFactory and associated Unit Tests
- Added helper function to `coreWrapper`
- Updated SetToken unit tests and use of `DetailedERC20`
- Removed setRegistry and associated migrations, tests, etc.


contracts/core/SetTokenFactory.sol
```diff
@@ -1,18 +1,58 @@
...
+        uint _naturalUnit,
+        string _name,
+        string _symbol
     ) 
-        public
+        external
+        onlyAuthorized
         returns
-        (address) 
+        (address /* setToken address */)
```
asoong: Seems strange, I think we should just consistently be explicit about the return value in the javadocs

test/setToken.spec.ts
```diff
@@ -114,42 +112,28 @@ contract("SetToken", (accounts) => {
...
+        { from: caller },
+      );
+    }
+
     describe(`of Standard Set`, () => {
       beforeEach(async () => {
         await resetAndDeployComponents(2);
       });
 
       it("should work with the correct data", async () => {
```
asoong: Ideally these are separate test cases, be explicit about what you're testing, or if you're grouping them together.
`it creates a set with the correct symbol`
`it creates a set ...`

test/setToken.spec.ts
```diff
@@ -114,42 +112,28 @@ contract("SetToken", (accounts) => {
...
-        const setTokenTruffleInstance = await SetToken.new(
-          _.map(components, (component) => component.address),
-          units,
-          STANDARD_NATURAL_UNIT,
-          TX_DEFAULTS,
-        );
+
+        const setTokenTruffleInstance = await subject();
```
asoong: This setup, along with setup directly below it should go into coreWrapper (or some other util). 

```
const setTokenWeb3Instance = web3.eth
             .contract(setTokenTruffleInstance.abi)
             .at(setTokenTruffleInstance.address);
 
const setTokenInstance: SetTokenContract = new SetTokenContract(setTokenWeb3Instance, TX_DEFAULTS);
```

asoong: The subject can just call `coreWrapper.deploySetTokenAsync(...)` and toggle those for nested test cases.

test/setToken.spec.ts
```diff
@@ -191,24 +184,28 @@ contract("SetToken", (accounts) => {
 
       it("should not work with mismatched quantity of units and tokens", async () => {
```
asoong: Put the context into a describe block. https://docs.google.com/document/d/1_eBX4js6XkrZdVM3gEsiufYz0cn-Wu4TfAG0Av5bgUw/edit#heading=h.teczyosuxzcd

describe("when there is a mismatch in quantity of units and components"...
beforeEach( () => units.pop)

See other spec files for examples.

test/setTokenFactory.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  after(async () => {
+    ABIDecoder.removeABI(SetTokenFactory.abi);
+  });
+
+  describe("#create", async () => {
+    let caller: Address = authorizedAccount;
+    let components: Address[] = [];
+    let units: BigNumber[] = [];
+    let naturalUnit: BigNumber = new BigNumber(0);
```
asoong: There is a ZERO constant now

test/setTokenFactory.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    let name: string = "";
```
asoong: It doesn't seem like you're manipulating these in nested test cases, just add it directly to the subject? Just give a default name/symbol instead of empty string.

test/setTokenFactory.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      });
+
+      it("should create a SetToken correctly", async () => {
+        const newSetAddress = await subject();
+
+        expect(newSetAddress).to.not.be.null;
+      });
+
+      describe("when the caller is not authorized", async () => {
+        before(async () => {
```
asoong: beforeEach

test/setToken.spec.ts
felix2feng: Mmm. I'm not going to do a full refactor of setToken in this PR. Whoever is working on setToken should be doing so.

PR #49, opened 2018-06-13
Add Redeem functionality to the Core contract

justinkchen:
Just pushing up to get logic checked out.

Still need to add specs for Core#redeem and SetToken#burn

@SetProtocol/set-protocol-core

### DIFF
--- a/contracts/core/Core.sol
+++ b/contracts/core/Core.sol
@@ -426,24 +426,15 @@ contract Core is
         uint _quantity
     )
         public
-        isValidSet(_tokenAddress)
+        isValidSet(_setAddress)
         isNonZero(_quantity)
+        isNaturalUnitMultiple(_quantity, _setAddress)
     {
-        // Check if multiple of naturalUnit
-        uint naturalUnit = ISetToken(_tokenAddress).naturalUnit();
-        require(isMultipleOfNaturalUnit(_quantity, naturalUnit));
-
-        // Burn the Set token (thereby decrementing the SetToken balance)
-        ISetToken(_tokenAddress).burn(msg.sender, _quantity);
+        uint naturalUnit = ISetToken(_setAddress).naturalUnit();
 
         // Transfer the underlying tokens to the corresponding token balances
-<<<<<<< HEAD
-        address[] memory components = ISetToken(_tokenAddress).getComponents();
-        uint[] memory units = ISetToken(_tokenAddress).getUnits();
-=======
-        address[] memory components = SetToken(_setAddress).getComponents();
-        uint[] memory units = SetToken(_setAddress).getUnits();
->>>>>>> Fix variable name
+        address[] memory components = ISetToken(_setAddress).getComponents();
+        uint[] memory units = ISetToken(_setAddress).getUnits();
         for (uint16 i = 0; i < components.length; i++) {
             address currentComponent = components[i];
             uint currentUnit = units[i];
@@ -468,6 +459,9 @@ contract Core is
                 tokenValue
             );
         }
+
+        // Burn the Set token (thereby decrementing the SetToken balance)
+        ISetToken(_setAddress).burn(msg.sender, _quantity);
     }
 
     /* ============ Private Functions ============ */
@@ -489,17 +483,6 @@ contract Core is
         returns(uint)
     {
         return _quantity.div(_naturalUnit).mul(_componentUnits);
-    }
-
-    function isMultipleOfNaturalUnit(
-        uint _quantity,
-        uint _naturalUnit
-    )
-        pure
-        internal
-        returns(bool)
-    {
-        return (_quantity % _naturalUnit) == 0;
     }
 
 }
--- a/contracts/core/Core.sol
+++ b/contracts/core/Core.sol
@@ -430,6 +430,9 @@ contract Core is
         isNonZero(_quantity)
         isNaturalUnitMultiple(_quantity, _setAddress)
     {
+        // Burn the Set token (thereby decrementing the SetToken balance)
+        ISetToken(_setAddress).burn(msg.sender, _quantity);
+
         uint naturalUnit = ISetToken(_setAddress).naturalUnit();
 
         // Transfer the underlying tokens to the corresponding token balances
@@ -459,9 +462,6 @@ contract Core is
                 tokenValue
             );
         }
-
-        // Burn the Set token (thereby decrementing the SetToken balance)
-        ISetToken(_setAddress).burn(msg.sender, _quantity);
     }
 
     /* ============ Private Functions ============ */
--- a/contracts/core/Core.sol
+++ b/contracts/core/Core.sol
@@ -437,8 +437,13 @@ contract Core is
         ISetToken(_tokenAddress).burn(msg.sender, _quantity);
 
         // Transfer the underlying tokens to the corresponding token balances
+<<<<<<< HEAD
         address[] memory components = ISetToken(_tokenAddress).getComponents();
         uint[] memory units = ISetToken(_tokenAddress).getUnits();
+=======
+        address[] memory components = SetToken(_setAddress).getComponents();
+        uint[] memory units = SetToken(_setAddress).getUnits();
+>>>>>>> Fix variable name
         for (uint16 i = 0; i < components.length; i++) {
             address currentComponent = components[i];
             uint currentUnit = units[i];
--- a/contracts/core/Core.sol
+++ b/contracts/core/Core.sol
@@ -482,7 +482,7 @@ contract Core is
         uint _quantity,
         uint _naturalUnit
     )
-        view
+        pure
         internal
         returns(bool)
     {
--- a/contracts/core/Core.sol
+++ b/contracts/core/Core.sol
@@ -418,11 +418,11 @@ contract Core is
     /**
      * Function to convert Set Tokens into underlying components
      *
-     * @param _tokenAddress    The address of the Set token
+     * @param _setAddress    The address of the Set token
      * @param _quantity        The number of tokens to redeem
      */
     function redeem(
-        address _tokenAddress,
+        address _setAddress,
         uint _quantity
     )
         public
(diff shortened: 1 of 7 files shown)


contracts/core/Core.sol
```diff
@@ -59,6 +65,15 @@ contract Core is
         _;
     }
 
+    // Checks to make sure a valid token address is given.
+    modifier isValidTokenAddress(address _tokenAddress) {
```
felix2feng: I'm currently adding logic that allows checks to see whether the Set is Valid

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
             _quantity
         );
     }
+
+    /**
+     * Function to convert Set Tokens into underlying components
+     *
+     * @param _tokenAddress    The address of the Set token
```
felix2feng: Can we converge on whether we want to include types in the JavaDocs?

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
...
+     */
+    function redeem(
+        address _tokenAddress,
+        uint _quantity
+    )
+        public
+        isValidTokenAddress(_tokenAddress)
+        isNonZero(_quantity)
+    {
+        // Check if multiple of naturalUnit
```
felix2feng: We can make this check a private function, since it'll be used in redeem too

contracts/core/Core.sol
```diff
@@ -59,6 +65,15 @@ contract Core is
         _;
     }
 
+    // Checks to make sure a valid token address is given.
+    modifier isValidTokenAddress(address _tokenAddress) {
+        // Confirm address is not null
+        require(_tokenAddress != address(0));
+        // Confirm address is not Core address
+        require(_tokenAddress != address(this));
```
asoong: Core will either need to hold a mapping of valid set tokens or query factory for it. Instead of checking that the address is not `this`, it should just check that mapping.

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
...
+            uint currentUnits = units[i];
+
+            uint preTransferBalance = ERC20(currentComponent).balanceOf(msg.sender);
+
+            uint transferValue = calculateTransferValue(
+                currentUnits,
+                naturalUnit,
+                _quantity
+            );
+            require(ERC20(currentComponent).transfer(msg.sender, transferValue));
```
felix2feng: There shouldn't be any transferring in redeem. Just decrementing / incrementing balances of the vault

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
...
+        // Transfer the underlying tokens to the corresponding token balances
+        address[] memory components = SetToken(_tokenAddress).getComponents();
+        uint[] memory units = SetToken(_tokenAddress).getUnits();
+        for (uint16 i = 0; i < components.length; i++) {
+            address currentComponent = components[i];
+            uint currentUnits = units[i];
+
+            uint preTransferBalance = ERC20(currentComponent).balanceOf(msg.sender);
+
+            uint transferValue = calculateTransferValue(
```
felix2feng: This should just be tokenValue or something

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
...
+                currentUnits,
+                naturalUnit,
+                _quantity
+            );
+            require(ERC20(currentComponent).transfer(msg.sender, transferValue));
+
+            // Check that preTransferBalance + transfer value is the same as postTransferBalance
```
felix2feng: Don't need this check until we're actually moving tokens

contracts/core/Core.sol
```diff
@@ -219,4 +234,87 @@ contract Core is
...
+    {
+        // Check if multiple of naturalUnit
+        uint naturalUnit = SetToken(_tokenAddress).naturalUnit();
+        require(isMultipleOfNaturalUnit(_quantity, naturalUnit));
+
+        // Burn the Set token
+        SetToken(_tokenAddress).burn(_quantity);
+
+        // Decrement the Set Token amount (this will validate if balance has enough)
+        Vault(vaultAddress).decrementTokenOwner(
```
bweick: Set Token balances aren't kept in Vault but in the Set Token contract, burning the token changes the Set Token balance. Don't need this line of code.

contracts/core/Core.sol
justinkchen: let me know when this is in and i can refactor

contracts/core/Core.sol
justinkchen: seems like it can be handled by @felix2feng 's change

contracts/core/Core.sol
justinkchen: isn't it already`internal`?

contracts/core/Core.sol
justinkchen: ah okay so need to just make decrements to the component values in Vault

contracts/core/Core.sol
justinkchen: @felix2feng do we have any incrementing of the vault here? if we're redeeming, its just decrement right?

contracts/core/Core.sol
bweick: We should be incrementing the components in the Vault and decrementing (through burn) the Set Token in the SetToken contract if that makes sense. It's basically debiting one account (SetToken) and crediting another (Vault). Purely internal accounting, then if someone wants to withdraw their tokens they could call withdraw in Core and thats how they'd get the constituent tokens back.

justinkchen: For SetToken.burn...it currently burns from the `msg.sender` balance. do you know if that is going to be `Core.sol` in this case or the `msg.sender` who first called the `redeem` method?

bweick: It would be msg.sender because the function is being called in Core, so the parameter that is being passed is the msg.sender address as it related to Core. Inside the SetToken contract msg.sender should not be referenced

bweick: so the burn function should have another parameter that passes on the address to burn the token from (in this case msg.sender). Make sense?

contracts/core/Core.sol
justinkchen: this is the thing i was asking about @felix2feng

contracts/core/Core.sol
```diff
@@ -416,6 +415,55 @@ contract Core is
...
+            // Increment the component amount
+            IVault(vaultAddress).incrementTokenOwner(
+                msg.sender,
+                currentComponent,
+                tokenValue
+            );
+        }
+
+        // Burn the Set token (thereby decrementing the SetToken balance)
+        ISetToken(_setAddress).burn(msg.sender, _quantity);
```
felix2feng: We should burn the token first (not last) if we are doing conservative / optimistic accounting

contracts/core/SetToken.sol
```diff
@@ -211,13 +211,17 @@ contract SetToken is
...
+        isCore
+        isNonZero(_quantity)
     {
-        balances[msg.sender] = balances[msg.sender].sub(quantity);
-        totalSupply_ = totalSupply_.sub(quantity);
-        emit Transfer(msg.sender, address(0), quantity);
+        require(balances[_from] >= _quantity);
+
+        balances[_from] = balances[_from].sub(_quantity);
+        totalSupply_ = totalSupply_.sub(_quantity);
```
felix2feng: We should add an event for Transfer from the user to a nonexistent address.
```
emit Transfer(from, address(0), amount);
```

```

File: test/core.spec.ts [hunk 11]
```diff
@@ -896,6 +896,130 @@ contract("Core", (accounts) => {
...
+    let setToken: SetTokenContract;
+
+    beforeEach(async () => {
+      await deployCoreAndInitializeDependencies();
+
+      components = await coreWrapper.deployTokensAsync(2, ownerAccount);
+      await coreWrapper.approveTransfersAsync(components, transferProxy.address);
+
+      const componentAddresses = _.map(components, (token) => token.address);
+      componentUnits = _.map(components, () => ether(4)); // Multiple of naturalUnit
```
felix2feng: The component units dont have to be a multiple of naturalUnit. Only the input quantity does

test/core.spec.ts
```diff
@@ -896,6 +896,130 @@ contract("Core", (accounts) => {
...
+        componentUnits,
+        naturalUnit,
+        "Set Token",
+        "SET",
+      );
+
+      await coreWrapper.issueSetTokenAsync(
+        core,
+        setToken.address,
+        ether(2),
```
felix2feng: Maybe make this a variable

test/core.spec.ts
```diff
@@ -896,6 +896,130 @@ contract("Core", (accounts) => {
...
+    });
+
+    async function subject(): Promise<string> {
+      return core.redeem.sendTransactionAsync(
+        subjectSetToRedeem,
+        subjectQuantityToRedeem,
+        { from: ownerAccount },
+      );
+    }
+
```
felix2feng: Add a test for decremented `setToken` balance of the user, because it got burned

contracts/core/Core.sol
```diff
@@ -416,6 +415,55 @@ contract Core is
...
+    }
```
felix2feng: I wonder if its useful to emit a `Redeem` event.

asoong: +1. I didn't emit any events for issue, though now that I think of it, it might be useful to emit an event for when a Transfer was required for a token.

test/core.spec.ts
asoong: I did this in my test for issue so that I could test what happens if the unit is not a multiple of the natural unit (3)

contracts/core/Core.sol
justinkchen: Is it enough for the `emit Transfer(_from, address(0), _quantity);` event in `burn`?

test/core.spec.ts
justinkchen: I added that in the `setToken.spec.ts`

felix2feng: Hmm is there a way to validate that burn still got called though at the right place?

contracts/core/SetToken.sol
justinkchen: added

test/core.spec.ts
justinkchen: added spec

PR #51, opened 2018-06-14
Set Token mint

asoong:
Using the mechanism we described today to validate sender of a mint request. Justin, you should be using this in burn.

### DIFF
--- a/contracts/core/SetToken.sol
+++ b/contracts/core/SetToken.sol
@@ -21,6 +21,7 @@ import { DetailedERC20 } from "zeppelin-solidity/contracts/token/ERC20/DetailedE
 import { ERC20 } from "zeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 import { StandardToken } from "zeppelin-solidity/contracts/token/ERC20/StandardToken.sol";
 import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
+import { ISetFactory } from "./interfaces/ISetFactory.sol";
 import "../lib/AddressArrayUtils.sol";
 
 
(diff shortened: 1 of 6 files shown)


contracts/core/SetToken.sol
```diff
@@ -171,19 +184,31 @@ contract SetToken is
...
         external
+        isAuthorized
     {
-        balances[msg.sender] = balances[msg.sender].add(quantity);
-        totalSupply_ = totalSupply_.add(quantity);
-        emit Transfer(address(0), msg.sender, quantity);
+        // Update token balances
+        balances[_issuer] = balances[_issuer].add(_quantity);
+
+        // Add t
```
asoong: I"ll clean this up, ignore in your review.

contracts/core/SetToken.sol
```diff
@@ -61,13 +63,24 @@ contract SetToken is
...
+
     /* ============ Modifiers ============ */
 
     modifier isMultipleOfNaturalUnit(uint _quantity) {
         require((_quantity % naturalUnit) == 0);
         _;
     }
 
+    modifier isAuthorized() {
+        require(
```
felix2feng: Maybe add a comment that only core is allowed. Or change the modifier name something like `isCore`

contracts/core/SetTokenFactory.sol
```diff
@@ -30,6 +31,29 @@ import { Authorizable } from "../lib/Authorizable.sol";
...
+
+    // Address of the Core contract
+    address public core;
+
+    /* ============ No Constructor ============ */
+
+    /* ============ Setter Functions ============ */
+
+    /**
+     * Set core. Can only be set by owner of Core.
```
felix2feng: only be set by SetTokenFactory's owner.

PR #62, opened 2018-06-18
Brian/core internal

asoong:


test/utils/coreWrapper.ts
```diff
@@ -3,6 +3,7 @@ import * as _ from "lodash";
 import { AuthorizableContract } from "../../types/generated/authorizable";
 import { BadTokenMockContract } from "../../types/generated/bad_token_mock";
 import { CoreContract } from "../../types/generated/core";
+import { CoreInternalContract } from "../../types/generated/core_internal";
```
bweick: We could get rid of this

test/utils/coreWrapper.ts
```diff
@@ -29,6 +30,7 @@ import { randomIntegerLessThan } from "../utils/math";
 const Authorizable = artifacts.require("Authorizable");
 const BadTokenMock = artifacts.require("BadTokenMock");
 const Core = artifacts.require("Core");
+const CoreInternal = artifacts.require("CoreInternal");
```
bweick: ...and this

contracts/core/Core.sol
```diff
@@ -33,7 +34,8 @@ import { IVault } from "./interfaces/IVault.sol";
  * creating Sets, as well as all collateral flows throughout the system.
  */
 contract Core is
-    Ownable
+    CoreState,
```
justinkchen: is this necessary? `CoreInternal` also inherits from it so won't it be enough to just say `is CoreInternal`?

contracts/core/Core.sol
```diff
@@ -16,13 +16,14 @@
...
-
-import { Ownable } from "zeppelin-solidity/contracts/ownership/Ownable.sol";
-import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
+import { CoreInternal } from "./CoreInternal.sol";
+import { CoreState } from "./lib/CoreState.sol";
 import { ISetFactory } from "./interfaces/ISetFactory.sol";
 import { ISetToken } from "./interfaces/ISetToken.sol";
 import { ITransferProxy } from "./interfaces/ITransferProxy.sol";
 import { IVault } from "./interfaces/IVault.sol";
+import { Ownable } from "zeppelin-solidity/contracts/ownership/Ownable.sol";
```
justinkchen: nit: the package paths (i.e. non-relative paths) should remain at top of file as a general style guide

asoong: +1

contracts/core/Core.sol
asoong: +1

contracts/core/lib/CoreState.sol
```diff
@@ -0,0 +1,82 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+pragma solidity 0.4.24;
+
+
+/**
+ * @title CoreSate
```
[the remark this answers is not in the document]
asoong: +1

contracts/core/CoreInternal.sol
```diff
@@ -0,0 +1,107 @@
...
+        external
+        onlyOwner
+    {
+        state.validFactories[_factoryAddress] = false;
+    }
+
+    /**
+     * Disable a set token in the mapping of tracked set tokens.
+     *
+     * @param  _setAddress   The address of the SetTokenFactory to remove
```
justinkchen: Fix comment here to be `SetToken`

asoong: +1

test/coreInternal.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+    it("sets vault address correctly", async () => {
+      await subject();
+
+      const storedVaultAddress = await core.vaultAddress.callAsync();
+      expect(storedVaultAddress).to.eql(vault.address);
+    });
+
+    describe("when the caller is not the owner of the contract", async () => {
+      beforeEach(async () => {
+        caller = otherAccount;
```
justinkchen: the way that this is currently setup, the `caller` will be `otherAccount` after this test is run instead of being reset to `ownerAccount` like it should be?

test/coreInternal.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+    it("sets vault address correctly", async () => {
+      await subject();
+
+      const storedVaultAddress = await core.vaultAddress.callAsync();
+      expect(storedVaultAddress).to.eql(vault.address);
+    });
+
+    describe("when the caller is not the owner of the contract", async () => {
+      beforeEach(async () => {
+        subjectCaller = otherAccount;
```
justinkchen: the way this is set up, `subjectCaller` will be `otherAccount` after this test is run instead of being reset to `ownerAccount`.  If a later test needed `subjectCaller` to be `ownerAccount`, it would fail

asoong: +1

PR #65, opened 2018-06-21
Split up Create function into its own Extension

felix2feng:
- Move Create into its own extension
- Update specs


contracts/core/extensions/CoreCreate.sol
```diff
@@ -0,0 +1,111 @@
...
+
+import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
+import { CoreState } from "../lib/CoreState.sol";
+import { ISetFactory } from "../interfaces/ISetFactory.sol";
+
+
+/**
+ * @title Core Create
+ * @author Set Protocol
+ *x
```
asoong: ?

PR #66, opened 2018-06-21
Core accounting extension

asoong:


contracts/core/extensions/CoreAccounting.sol
```diff
@@ -0,0 +1,179 @@
...
+import { CoreState } from "../lib/CoreState.sol";
+import { ITransferProxy } from "../interfaces/ITransferProxy.sol";
+import { IVault } from "../interfaces/IVault.sol";
+
+
+/**
+ * @title Core Internal
+ * @author Set Protocol
+ *
+ * The CoreInternal contract contains methods to alter state that tracks contract
```
justinkchen: update comment

PR #68, opened 2018-06-24
Brian/core interface

bweick:


justinkchen:
I reran your circleci to get it fixed

contracts/core/extensions/CoreFactory.sol
```diff
@@ -80,7 +80,7 @@ contract CoreFactory is
         string _name,
         string _symbol
     )
-        public
+        external
```
justinkchen: do we want to move this under another header? Right now its under `/* ============ Public Functions ============ */`

contracts/core/interfaces/ICore.sol
```diff
@@ -0,0 +1,175 @@
...
+    function disableSet(
+        address _setAddress
+    )
+        external;
+
+    /**
+     * Issue
+     *
+     * @param  _setAddress   Address of set to issue
+     * @param  _quantity     Quantity of set to issue
```
felix2feng: We should add that the quantity to issue should be a multiple of the naturalUnit of the Set

PR #78, opened 2018-07-02
Alex/decode order data

asoong:
- Updated structs, removing overall order data header
- Remove previously implemented `MockExchangeOrderHandler` which was added just to test the `ExchangeOrderHandler` library
- Parse header as part of fillOrder
- Add `ExchangeDispatcher` as an Core extension to manage valid exchanges that can be used to fill the issuanceOrder


contracts/external/LibBytes.sol
```diff
@@ -179,7 +179,7 @@ library LibBytes {
             "FROM_LESS_THAN_TO_REQUIRED"
         );
         require(
-            to < b.length,
+            to <= b.length,
```
asoong: @felix2feng did they respond to your question about this?

test/utils/orderWrapper.ts
```diff
@@ -4,46 +4,46 @@ import * as ethUtil from "ethereumjs-util";
...
-  return ethUtil.bufferToHex(
-    ethUtil.setLengthLeft(
-      ethUtil.toBuffer(orderCount), 32)
-    );
+  const exchangeOrderDatum: Buffer[] = [];
+  _.times(orderCount, (index) => {
+    const exchange = _.sample(EXCHANGES);
+    exchangeOrderDatum.push(paddedBufferForData(exchange));
+
+    const orderLength = _.random(120, 160)
```
asoong: Hoping to sub any kind of util function we write later that generates a real exchange order that mirrors what `setProtocol.js` will return

contracts/core/extensions/CoreExchangeDispatcher.sol
```diff
@@ -0,0 +1,67 @@
...
+    limitations under the License.
+*/
+
+pragma solidity 0.4.24;
+
+import { Ownable } from "zeppelin-solidity/contracts/ownership/Ownable.sol";
+import { CoreState } from "../lib/CoreState.sol";
+
+
+// /**
```
felix2feng: Why are there 2 types of comments here?

contracts/core/lib/CoreState.sol
```diff
@@ -29,6 +29,9 @@ contract CoreState {
     /* ============ Structs ============ */
 
     struct State {
+        // Mapping of exchagne enumeration to address
```
felix2feng: "exchange"

contracts/core/lib/ExchangeHandler.sol
```diff
@@ -0,0 +1,63 @@
...
+ * @author Set Protocol
+ *
+ * This library contains functions and structs to assist with parsing exchange orders data
+ */
+library ExchangeHandler {
+
+    // ============ Structs ============
+
+    struct OrderHeader {
+        uint8 exchange;
```
felix2feng: FYI - Currently, we don't get any storage savings from packing with just a `uint8`

PR #79, opened 2018-07-02
Brian/validate order sig

bweick:


contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +36,131 @@ contract CoreIssuanceOrder is
...
+            order.setAddress,
+            _fillQuantity
+        );
+    }
+
+    function generateOrderHash(
+        address[4] _addresses,
+        uint[5] _values
+    )
+        public
```
asoong: Public? This seems like something that should be internal or private

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +36,131 @@ contract CoreIssuanceOrder is
...
+        address makerAddress;
+        address makerToken;
+        uint256 makerTokenAmount;
+        uint256 expiration;
+        address relayerToken;
+        uint256 relayerTokenAmount;
+        uint256 salt;
+        bytes32 orderHash;
+    }
+
```
asoong: Add some descriptions including what’s at each index of the array params?

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +36,131 @@ contract CoreIssuanceOrder is
...
+        );
+    }
+
+    function validateOrder(
+        IssuanceOrder _order,
+        uint _fillQuantity
+    )
+        internal
+    {
+        require(_order.makerTokenAmount > 0 && _order.quantity > 0);
```
asoong: Error message?

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +36,131 @@ contract CoreIssuanceOrder is
...
+            _fillQuantity
+        );
+    }
+
+    function generateOrderHash(
+        address[4] _addresses,
+        uint[5] _values
+    )
+        public
+        returns(bytes32)
```
asoong: Same

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +36,131 @@ contract CoreIssuanceOrder is
...
         //Issue Set
-        issueInternal(_maker, _setAddress, _quantity);
+        issueInternal(
+            order.makerAddress,
+            order.setAddress,
+            _fillQuantity
+        );
+    }
+
+    function generateOrderHash(
```
asoong: Same, docs explaining these params

contracts/core/extensions/CoreIssuanceOrder.sol
bweick: Yeah this one I struggled with, there's no harm in making it public but makes more sense for it to be internal. For expediency purposes I just made it public so that it could be tested. We haven't come up with a framework for testing internal functions I feel like.

contracts/core/extensions/CoreIssuanceOrder.sol
bweick: ?

contracts/core/extensions/CoreIssuanceOrder.sol
bweick: Will do

contracts/core/extensions/CoreIssuanceOrder.sol
bweick: Yeah needed to add javadocs and comments

contracts/core/extensions/CoreIssuanceOrder.sol
asoong: I think these belong in a library of some sort as they don’t alter state. You could then test them by adding a mock contract, kind of how we had one for ExchangeOrderHandler

contracts/core/extensions/CoreIssuanceOrder.sol
asoong: Same as in, looks like it should be either internal or belongs to some kind of IssuanceOrder library

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -36,17 +37,105 @@ contract CoreIssuanceOrder is
...
+    )
+        internal
+        view
+    {
+        // Make sure makerTokenAmount and Set Token to issue is greater than 0.
+        require(
+            _order.makerTokenAmount > 0 && _order.quantity > 0,
+            INVALID_TOKEN_AMOUNTS
+        );
+        // Make sure the order hasn't expired
```
asoong: nit: space when you work on this next

contracts/core/lib/OrderLibrary.sol
```diff
@@ -0,0 +1,114 @@
...
+        recAddress = ecrecover(
+            keccak256(abi.encodePacked(msgPrefix, _orderHash)),
+            _v,
+            _r,
+            _s
+        );
+
+        return recAddress == _signerAddress;
+    }
+
```
asoong: nit: space

PR #85, opened 2018-07-03
Brian/factory and token arrays

bweick:


test/core/extensions/coreInternal.spec.ts
```diff
@@ -241,6 +255,13 @@ contract("CoreInternal", (accounts) => {
       expect(isSetValid).to.be.false;
     });
 
+    it("removes setToken address from setTokens array", async () => {
+      await subject();
+
+      const setTokens = await core.setTokens.callAsync();
+      expect(setTokens).to.not.include(setToken.address);
```
felix2feng: You can also add an assertion to expected array length

contracts/core/extensions/CoreInternal.sol
```diff
@@ -92,6 +93,13 @@ contract CoreInternal is
         isValidFactory(_factoryAddress)
     {
         state.validFactories[_factoryAddress] = false;
+        for (uint256 i = 0; i < state.factories.length; i++) {
```
felix2feng: Can we create an address[] helper method for this?

bweick: I'll add to Trello and we can decide. Wanna get this merged within the scope of what it is.

PR #86, opened 2018-07-03
Redeem and withdraw

asoong:
Purposely does not call `withdraw` as it results in redundant transfer value calculations and unnecessary updating of vault balances. Therefore, some of the tests may seem redundant and similar to `#redeem`.

### DIFF
--- a/contracts/core/extensions/CoreIssuance.sol
+++ b/contracts/core/extensions/CoreIssuance.sol
@@ -91,10 +91,9 @@ contract CoreIssuance is
         uint[] memory units = ISetToken(_setAddress).getUnits();
         for (uint16 i = 0; i < components.length; i++) {
             address currentComponent = components[i];
-            uint currentUnit = units[i];
 
             uint tokenValue = calculateTransferValue(
-                currentUnit,
+                units[i],
                 naturalUnit,
                 _quantity
             );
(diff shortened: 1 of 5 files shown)


test/core/extensions/coreIssuance.spec.ts
```diff
@@ -475,4 +475,167 @@ contract("CoreIssuance", (accounts) => {
...
+
+      const expectedNewBalances = _.map(existingTokenBalances, (balance, idx) => {
+        const quantityToRedeem = subjectQuantityToRedeem.div(naturalUnit).mul(componentUnits[idx]);
+        return balance.add(quantityToRedeem);
+      });
+      const newTokenBalances = await erc20Wrapper.getTokenBalances(components, ownerAccount);
+      expect(newTokenBalances).to.eql(expectedNewBalances);
+    });
+
+    describe("when the withdraw mask includes one component", async () => {
```
bweick: Is it worth it to also include a check to make doubly sure that no component tokens have been transferred to the user that weren't supposed to be transferred, purely to prevent any potential regressions. I don't feel overly strongly...could be over testing.

bweick:
Yeah it checks whether we’ve done our internal accounting right but it doesn’t check to make sure those tokens didn’t accidentally get transferred out of our system. It wouldn’t break with the current code but could with a bad regression. Like I said...could be over testing. Conversely there could be a test to make sure the withdrawn tokens weren’t accidentally incremented in the vault.

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -91,10 +93,9 @@ contract CoreIssuance is
         uint[] memory units = ISetToken(_setAddress).getUnits();
         for (uint16 i = 0; i < components.length; i++) {
             address currentComponent = components[i];
-            uint currentUnit = units[i];
 
             uint tokenValue = calculateTransferValue(
-                currentUnit,
+                 units[i],
```
justinkchen: extranous space here it seems

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -115,6 +116,75 @@ contract CoreIssuance is
...
+        ISetToken(_setAddress).burn(msg.sender, _quantity);
+
+        // Fetch Set token properties
+        uint naturalUnit = ISetToken(_setAddress).naturalUnit();
+        address[] memory components = ISetToken(_setAddress).getComponents();
+        uint[] memory units = ISetToken(_setAddress).getUnits();
+
+        // Loop through and decrement vault balances for the set, withdrawing if requested
+        for (uint i = 0; i < components.length; i++) {
+
```
justinkchen: remove extra line

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -115,6 +116,75 @@ contract CoreIssuance is
...
+            // Calculate quantity to transfer
+            uint componentQuantity = calculateTransferValue(
+                 units[i],
```
justinkchen: another extraneous space here

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -115,6 +116,75 @@ contract CoreIssuance is
...
+
+            // Decrement the Set amount
+            IVault(state.vaultAddress).decrementTokenOwner(
+                _setAddress,
+                components[i],
+                componentQuantity
+            );
+
+            // Calculate bit index of current component
+            uint componentBitIndex = (2 ** i);
```
justinkchen: no need for parens?

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -115,6 +116,75 @@ contract CoreIssuance is
...
+        for (uint i = 0; i < components.length; i++) {
+
+            // Calculate quantity to transfer
+            uint componentQuantity = calculateTransferValue(
+                 units[i],
+                naturalUnit,
+                _quantity
+            );
+
+            // Decrement the Set amount
```
justinkchen: we're decrementing the `component` amount right? not the `Set` amount

contracts/core/extensions/CoreIssuance.sol
asoong: good eye

contracts/core/extensions/CoreIssuance.sol
asoong: +1

PR #90, opened 2018-07-05
Brian/open order amount

bweick:


test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -178,15 +249,15 @@ contract("CoreIssuanceOrder", (accounts) => {
 
     describe("when the set was not created through core", async () => {
       beforeEach(async () => {
-        issuanceOrderParams = await generateFillOrderParameters(NULL_ADDRESS, signerAddress, signerAddress, componentAddresses[0])
+        issuanceOrderParams = await generateFillOrderParameters(NULL_ADDRESS, signerAddress, signerAddress, componentAddresses[0], relayerAddress);
```
asoong: Why is the maker address you're passing in here also `signerAddress`?

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -196,29 +267,39 @@ contract("CoreIssuanceOrder", (accounts) => {
       });
     });
 
+    describe("when the order quantity is not a multiple of the natural unit of the set", async () => {
```
asoong: In these specs where the setup is all common, can you help identify what is is you're toggling? I would put `generateFillOrderParameters`. In `beforeEach` on line 124, you're already doing this generation, and then you're regenerating for each of these test cases. I'd create a variable for each one and then just toggle the specific variable in each test case.

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -93,16 +94,21 @@ contract("CoreIssuanceOrder", (accounts) => {
     let components: StandardTokenMockContract[] = [];
     let componentUnits: BigNumber[];
     let setToken: SetTokenContract;
```
asoong: Let's tighten these up, remove any variables here, also in `cancelOrder` for variables that don't need to be toggled. 

`naturalUnit`, `components`, `componentUnits`, `setToken`

test/core/lib/orderLibrary.spec.ts
```diff
@@ -58,16 +59,34 @@ contract("OrderLibrary", (accounts) => {
...
     let subjectMaker: Address;
     let signerAddress: Address;
+    let relayerAddress: Address;
+    let orderQuantity: BigNumber;
+    let makerTokenAmount: BigNumber;
+    let timeToExpiration: number;
 
     let issuanceOrderParams: any;
 
     beforeEach(async () => {
```
asoong: nit: space

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -225,26 +240,36 @@ contract CoreIssuanceOrder is
      * Validate order params are still valid
      *
      * @param  _order           IssuanceOrder object containing order params
-     * @param  _fillQuantity    Quantity of Set to be filled
+     * @param  _executeQuantity    Quantity of Set to be filled
      */
     function validateOrder(
         OrderLibrary.IssuanceOrder _order,
-        uint _fillQuantity
+        uint _executeQuantity
```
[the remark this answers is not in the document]
bweick: Si Papi

PR #96, opened 2018-07-07
Alex/common math

asoong:


contracts/mocks/core/exchange-wrappers/lib/ZeroExOrderDataHandlerMock.sol
```diff
@@ -1,13 +1,13 @@
 pragma solidity 0.4.24;
 pragma experimental "ABIEncoderV2";
 
-import { ZeroExOrderDataHandler } from "../../core/exchange-wrappers/lib/ZeroExOrderDataHandler.sol";
```
felix2feng: Absolute path could be helpful here.

test/contracts/lib/commonMath.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+BigNumberSetup.configure();
+ChaiSetup.configure();
+const { expect, assert } = chai;
+
+contract("CommonMathMock", (accounts) => {
+  const [ownerAccount] = accounts;
+
+  let commonMathLibrary: CommonMathMockContract;
+
+  describe("#testMaxUInt256", async () => {
```
felix2feng: Thanks for doing this

PR #97, opened 2018-07-07
Update list of artifacts to include some of the newer ones like Core …

justinkchen:
…while removing old deleted ones like AddressArrayUtils


types/base_contract.ts
```diff
@@ -2,6 +2,11 @@ import * as _ from "lodash";
 import * as Web3 from "web3";
 import { TxData, TxDataPayable } from "./common";
 
+export const CONTRACT_WRAPPER_ERRORS = {
```
inje: Are we using this?

artifacts/ts/Migrations.ts
```diff
@@ -65,14 +65,14 @@ export const Migrations =
...
-  "sourcePath": "/Users/alexsoong/Source/set-protocol/set-protocol-contracts/contracts/Migrations.sol",
+  "source": "pragma solidity 0.4.24;\n\n\ncontract Migrations {\n  address public owner;\n  uint public lastCompletedMigration;\n\n  modifier restricted() {\n    if (msg.sender == owner) _;\n  }\n\n  constructor() public {\n    owner = msg.sender;\n  }\n\n  function setCompleted(uint completed) public restricted {\n    lastCompletedMigration = completed;\n  }\n\n  function upgrade(address newAddress) public restricted {\n    Migrations upgraded = Migrations(newAddress);\n    upgraded.setCompleted(lastCompletedMigration);\n  }\n}\n",
+  "sourcePath": "/Users/justinkchen/workspace/set-protocol-contracts/contracts/Migrations.sol",
```
inje: I'm guessing this is autogenerated since Alex's name was in the last `"sourcePath"`. I'm guessing this is normal

artifacts/ts/IExchange.ts
```diff
@@ -0,0 +1,284 @@
...
+  "sourcePath": "/Users/justinkchen/workspace/set-protocol-contracts/contracts/core/interfaces/IExchange.sol",
```
inje: Seeing an absolute local path. Will this work?

PR #114, opened 2018-07-12
Add type definition module for solc

felix2feng:


types/modules.d.ts
```diff
@@ -23,3 +23,5 @@ declare module "abi-decoder" {
     export function decodeLogs(logs: any[]): DecodedLog[];
     export function decodeMethod(data: string): DecodedMethod;
 }
+
+declare module "sol-trace-set";
```
justinkchen: the official way is supposed to be to create a folder name `sol-trace-set` in `types` and add this line to the `index.d.ts`

justinkchen: same for `abi-decoder`

PR #115, opened 2018-07-12
Fixed minor issue with slice parameters

felix2feng:
- Was previously making incorrect assumptions about where in memory the slice was beginning and ending.


contracts/core/exchange-wrappers/ZeroExExchangeWrapper.sol
```diff
@@ -117,7 +117,7 @@ contract ZeroExExchangeWrapper
                 signature
             );
 
-        // Ensure the taker token is allowed to be approved to the TransferProxy
+        // // Ensure the taker token is allowed to be approved to the TransferProxy
```
justinkchen: Extra comment to remove

PR #118, opened 2018-07-15
Artifacts: Update core artifacts

inje:


artifacts/json/Core.json
```diff
@@ -975,12 +1019,12 @@
         "unitAlias": ""
       },
       {
-        "absolutePath": "/Users/justinkchen/workspace/set-protocol-contracts/contracts/core/extensions/CoreIssuanceOrder.sol",
+        "absolutePath": "/Users/inje/Documents/repos/set-protocol-contracts/contracts/core/extensions/CoreIssuanceOrder.sol",
```
felix2feng: I wonder if its weird that the absolute path is your own directory

inje: I wondered that too when Justin updated artifacts, but it had no side-effects. Before Justin's absolute path, it was Alex's, and our contract migrations run just fine.

PR #119, opened 2018-07-15
Brian+alex/integrate taker wallet

bweick:


test/contracts/core/extensions/coreIssuanceOrderScenarios.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+    takerWalletWrapper = await exchangeWrapper.deployTakerWalletExchangeWrapper(transferProxy);
+
+    // TODO: Move these authorizations into setDefaultStateAndAuthrorizations
+    await coreWrapper.addAuthorizationAsync(takerWalletWrapper, core.address);;
+    await coreWrapper.addAuthorizationAsync(transferProxy, takerWalletWrapper.address);
+
+    await coreWrapper.setDefaultStateAndAuthorizationsAsync(core, vault, transferProxy, setTokenFactory);
+  });
+
+  describe.only("#fillOrder", async () => {
```
felix2feng: .only!

contracts/core/exchange-wrappers/TakerWalletWrapper.sol
```diff
@@ -77,39 +78,31 @@ contract TakerWalletWrapper is
...
+            // Ensure allowance of transfer from this wrapper to TransferProxy
+            ERC20Wrapper.ensureAllowance(
                 takerToken,
+                address(this),
+                transferProxy,
                 takerTokenAmount
             );
 
+            // Record taker token and amount to return values
+            uint256 orderCount = scannedBytes >> 6;
```
felix2feng: How does this work? Does it increment one at a time?

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -278,13 +280,28 @@ contract CoreIssuanceOrder is
...
+            address[] memory componentFillTokens = new address[](header.orderCount);
+            uint[] memory componentFillAmounts = new uint[](header.orderCount);
+            (componentFillTokens, componentFillAmounts) = IExchange(exchange).exchange(
+                msg.sender,
+                header.orderCount,
+                bodyData
+            );
+
+            // Transfer component tokens from wrapper to vault
+            batchDepositInternal(
```
felix2feng: Wow that's really nice!

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -423,10 +440,15 @@ contract CoreIssuanceOrder is
                 _order.makerAddress,
                 _order.requiredComponents[i]
             );
-            //require(currentBal >= requiredBalances[i]);
+            require(currentBal >= requiredBalances[i]);
         }
 
-        settleAccounts(_order, _fillQuantity, requiredMakerTokenAmount, makerTokenAmountUsed);
+        settleAccounts(
```
felix2feng: Do we have the logic to send remaining makerTokens to the taker/arb yet?

bweick: That's in the settle accounts function itself

contracts/core/exchange-wrappers/TakerWalletWrapper.sol
asoong: Yes, but this is particular to our workflow we scan 64 bytes at at time, so you can think of it as counting from 0 starting on the 7th bit.

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -171,4 +164,63 @@ contract CoreAccounting is
...
+     * @param  _tokenAddress    The address of the ERC20 token
+     * @param  _quantity        The number of tokens to deposit
+     */
+    function depositInternal(
+        address _from,
+        address _to,
+        address _tokenAddress,
+        uint _quantity
+    )
+        public
```
justinkchen: shouldn't this be `internal`?

contracts/core/interfaces/ICoreAccounting.sol
```diff
@@ -0,0 +1,43 @@
...
+
+
+/**
+ * @title ICoreIssuance
+ * @author Set Protocol
+ *
+ * The ICoreIssuance Contract defines all the functions exposed in the CoreIssuance
+ * extension.
+ */
+contract ICoreAccounting {
```
justinkchen: It's a little weird to be using an abstract class here instead of `interface` (same for ICoreIssuance)...I think solidity throws warnings if methods in interfaces aren't `external` which wouldn't work here though but I wonder if we should be splitting up the naming of interfaces from abstract classes

bweick: Yeah we can't use internal functions in libraries so we have to define it as a contract. Which unfortunately makes us have to be more careful with inheritance when the Core inherits the IssuanceOrder extension lest we overwrite the actual logic of the batchDepositInternal function. If we want to name it something else we can do that since it isn't technically an interface thought it is intended to act as such.

contracts/core/extensions/CoreAccounting.sol
bweick: Yes...important catch. Imma push a fix immediately.

PR #125, opened 2018-07-20
Fix some linting, move utils out of test folder

justinkchen:
There's actually still a second round of linting that needs to occur but was going to leave that for separate PR.  This pulls the whole `utils` folder out of `tests` and adds it to the `dist/` folder so that setProtocol.js can also use it.


asoong:
LGTM

PR #127, opened 2018-07-21
Brian/full scenario analysis

bweick:


contracts/core/Vault.sol
```diff
@@ -127,7 +127,6 @@ contract Vault is
     )
         external
         onlyAuthorized
-        isNonZero(_quantity)
```
asoong: Why are these being removed?

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -211,7 +211,6 @@ contract CoreAccounting is
         uint[] _quantities
     )
         internal
-        isValidBatchTransaction(_tokenAddresses, _quantities)
```
asoong: Same ^

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -452,4 +452,24 @@ contract CoreIssuanceOrder is
         // Tally fill in orderFills mapping
         state.orderFills[_order.orderHash] = state.orderFills[_order.orderHash].add(_fillQuantity);
     }
+
+    function getPartialAmount(
```
asoong: Can this get moved into a lib

asoong:
Couple of questions, i'd like to see another parameter in each scenario that describes the outcome and/or flows that it goes through

contracts/core/Vault.sol
bweick: Because to we might batch deposit 0 tokens from one exchange but we still are passing on an array that contains the 0 so needed to remove that. Unless we want to do some array pruning somewhere.

contracts/core/extensions/CoreIssuanceOrder.sol
bweick: Sure could was looking for a lib that made sense in our set up so just left it as is. But could move into OrderLibrary I suppose.

contracts/core/Vault.sol
asoong: Can you give me an example? I think this is suboptimal for gas, prefer some kind of efficient pruning

contracts/core/extensions/CoreIssuanceOrder.sol
asoong: +1

contracts/core/extensions/CoreAccounting.sol
bweick: As I go back...I agree that there's no reason to remove those modifiers. It was breaking before and I thought that was the case...I wonder if something was changed in a rebase cause it definitely shouldn't change anything.

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -152,7 +152,7 @@ contract CoreIssuanceOrder is
         // Settle Order
         settleOrder(order, _fillQuantity, _orderData);
 
-        //Issue Set
+        //Issue Settle
```
justinkchen: need a space after `//`, i think the precommit linting should catch these things now though?

PR #131, opened 2018-07-26
Linked OrderLibrary to CoreIssuanceOrder. Some OrderLibrary fxns now external.

bweick:


utils/coreWrapper.ts
```diff
@@ -92,6 +93,11 @@ export class CoreWrapper {
   public async deployMockOrderLibAsync(
     from: Address = this._tokenOwnerAddress
   ): Promise<OrderLibraryMockContract> {
+    const truffleOrderLibrary = await OrderLibrary.new(
```
asoong: Can we be smarter about this? Ideally this is only deployed once per coreWrapper. Maybe save the instance of OrderLibrary as an instance variable on coreWrapper (like _tokenOwnerAddress)

PR #133, opened 2018-07-27
Alex/update utils

asoong:


test/contracts/core/extensions/coreExchangeDispatcher.spec.ts
```diff
@@ -1,6 +1,6 @@
-import * as chai from 'chai';
-
 import * as ABIDecoder from 'abi-decoder';
+import * as chai from 'chai';
+import { SetProtocolUtils as utils }  from 'set-protocol-utils';
```
inje: Is it best practice to change a class with PascalCase to a variable with camelCase?

asoong: I personally think it's more human readable

inje: I think so too. What I mean is should `utils` be `Utils` since it's a class

inje: The edge case is when you need to use static methods with the non-static methods.

I think a cleaner approach is to capitalize so we avoid something like this:
```
import { SetProtocolUtils as utils }  from 'set-protocol-utils';

const utilities = new utils(web3);
```
and instead have something like this
```
import { SetProtocolUtils as Utils }  from 'set-protocol-utils';

const utils = new Utils(web3);
```
Makes the instance delineated from the parent class.

Lmk if that makes sense.

PR #135, opened 2018-07-27
Brian/core optimizations

bweick:
Focuses on removing modifiers, error messages, and adding vars that house interfaces for Core. Overall cut down big time on bytes and gas. Dealing with the interfaces was a bit of a trade off where we add some bytes and deploy costs but reduce function call costs.


contracts/core/extensions/CoreAccounting.sol
```diff
@@ -215,8 +199,21 @@ contract CoreAccounting is
...
+
+        // Confirm an empty _quantities array is not passed
+        require(
+            _quantities.length > 0
+        );
+
+        // Confirm there is one quantity for every token address
+        require(
+            _tokens.length == _quantities.length
+        );
```
asoong: nit: whitespace

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -215,8 +199,21 @@ contract CoreAccounting is
         uint[] _quantities
     )
         internal
-        isValidBatchTransaction(_tokens, _quantities)
     {
+        // Confirm and empty _tokens array is not passed
+        require(
```
asoong: for style, since we are no longer passing in messages, let's have the check be all in one line

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -82,17 +92,33 @@ contract CoreIssuance is
         uint _quantity
     )
         external
-        isValidSet(_set)
-        isPositiveQuantity(_quantity)
-        isNaturalUnitMultiple(_quantity, _set)
     {
+        // Declare interface variables
+        ISetToken set = ISetToken(_set);
```
asoong: nit: setToken instead of set

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -139,17 +165,33 @@ contract CoreIssuance is
         uint _toWithdraw
     )
         external
-        isValidSet(_set)
-        isPositiveQuantity(_quantity)
```
asoong: Let's double check whether these checks are actually necessary, what do cases where these don't hold true expose us to?

asoong: One reason I ask is the data type is already uint.

test/contracts/core/extensions/coreAccounting.spec.ts
```diff
@@ -174,7 +174,9 @@ contract('CoreAccounting', accounts => {
     it('transfers the correct amount of tokens to the caller', async () => {
       const existingOwnerTokenBalance = await mockToken.balanceOf.callAsync(ownerAccount);
 
-      await subject();
+      const txHash = await subject();
```
asoong: remove?

PR #146, opened 2018-08-06
Add eth gas reporter

felix2feng:
Really cool that it adds gas usage after each unit test
image


asoong:
I wonder if we can make this run for tests only and not coverage

PR #148, opened 2018-08-07
redeemInVault functionality added with tests.

bweick:


contracts/core/extensions/CoreIssuance.sol
```diff
@@ -204,6 +204,73 @@ contract CoreIssuance is
...
+
+        // Decrement ownership of Set token in the vault
+        vault.decrementTokenOwner(
+            msg.sender,
+            _set,
+            _quantity
+        );
+
+        // Burn the Set token from the vault's amount
+        setToken.burn(
```
asoong: This is the only statement that's different from redeem. Can we combine the two functions and have one call the other? The public redeem must be called with msg.sender.

bweick: Well that and decrementing the sender's setToken balance in the vault. I feel like it's annoyingly different enough to have to have a full fxn. We could make everything below that a separate fxn.

PR #159, opened 2018-08-10
Add integration test with 0x orders

asoong:


contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -403,6 +403,9 @@ contract CoreIssuanceOrder is
             _order.makerAddress
         );
 
+        // Verify maker token used is less than amount allocated that user signed
+        // require(makerTokenAmountUsed <= requiredMakerTokenAmount);
```
asoong: @bweick Is this necessary? If makerTokenAmountUsed is greater, than the issuance order taker actually gets less. LMK if i'm misunderstanding something

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -242,22 +269,188 @@ contract('CoreIssuanceOrder', accounts => {
...
-        signerAccount,
-        subjectCaller,
-        makerToken.address,
-        relayerAccount,
-        relayerToken.address,
-        subjectQuantityToIssue,
-        ether(10),
-        ether(3),
-        issuanceOrderParams.orderHash,
+        setToken.address,              // setAddress
```
bweick: Since we have these comments here now...do we wanna just add the same ones to the log test for the partial fill scenario?

PR #167, opened 2018-08-14
[WIP] brian/rebalancing_set_basic

bweick:
First run implementation of rebalancing set. WIP. Testing not complete b/c need to find way to more gracefully move through states and need future functionality for full test.

### DIFF
--- a/contracts/core/RebalancingSet.sol
+++ b/contracts/core/RebalancingSet.sol
@@ -19,6 +19,7 @@ pragma solidity 0.4.24;
 import { DetailedERC20 } from "zeppelin-solidity/contracts/token/ERC20/DetailedERC20.sol";
 import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
 import { StandardToken } from "zeppelin-solidity/contracts/token/ERC20/StandardToken.sol";
+import { ISetFactory } from "./interfaces/ISetFactory.sol";
 
 /**
  * @title SetToken
@@ -70,6 +71,24 @@ contract RebalancingToken is
     uint256 public remainingCurrentSets;
     uint256 public rebalanceSetSupply;
 
+    /* ============ Events ============ */
+
+    event NewManagerAdded(
+        address newManager,
+        address oldManager
+    );
+
+    event RebalanceProposed(
+        address rebalancingSet,
+        address indexed auctionLibrary,
+        uint256 indexed proposalPeriodEndTime
+    );
+
+    event RebalanceStarted(
+        address oldSet,
+        address newSet
+    );
+
 
     /* ============ Constructor ============ */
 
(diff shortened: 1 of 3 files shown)


contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
+/*
```
felix2feng: Maybe we can move the SetToken, SetTokenFactory, RebalancingSet, and RebalancingSetFactory into their own directory

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+
+pragma solidity 0.4.24;
+
+import { DetailedERC20 } from "zeppelin-solidity/contracts/token/ERC20/DetailedERC20.sol";
+import { SafeMath } from "zeppelin-solidity/contracts/math/SafeMath.sol";
+import { StandardToken } from "zeppelin-solidity/contracts/token/ERC20/StandardToken.sol";
+import { ISetFactory } from "./interfaces/ISetFactory.sol";
+
+/**
+ * @title SetToken
```
felix2feng: Rebalancing Set Token

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+    event RebalanceFinished(
+        address newSet,
+        uint256 newShareRatio
+    );
+
+
+    /* ============ Constructor ============ */
+
+    /**
+     * Constructor function for Rebalancing Set Token
```
felix2feng: Perhaps we note that this initial Rebalancing Set uses a manager vs. other oracle system

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+     * @param _initialSet                   The initial set that collateralizes the Rebalancing set
+     * @param _initialShareRatio            How much of a Set (in gWei) equals one share
+     * @param _proposalPeriod               Amount of time for users to inspect a rebalance proposal
+     * @param _rebalanceInterval            The minimum amount of time between rebalances
+     * @param _name                         The Rebalancing Set's name
+     * @param _symbol                       The Rebalancing Set's symbol
+     */
+
+    constructor(
+        address _factory,
```
felix2feng: Do we want this Rebalancing Set to conform to a normal Set? So that it includes components, units, naturalUnit, etc.?

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+        uint256 _proposalPeriod,
+        uint256 _rebalanceInterval,
+        string _name,
+        string _symbol
+    )
+        public
+        DetailedERC20(_name, _symbol, 18)
+    {
+        // Require day long proposal period
+        require(_proposalPeriod > 86400);
```
felix2feng: I think ethereum has a global variable / unit called `day`

https://solidity.readthedocs.io/en/v0.4.24/units-and-global-variables.html?highlight=day

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+
+    /*
+     * Initiate rebalance for the rebalancing set. Users can now submit bids.
+     *
+     */
+    function rebalance()
+        external
+    {
+        // Must be in "Proposal" state before going into "Rebalance" state
+        require(rebalanceState == State.Proposal);
```
felix2feng: We could probably abstract the state transition and timing checks into a private function

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+        require(msg.sender == manager);
+
+        // New proposal cannot be made during a rebalance period
+        require(rebalanceState == State.Default);
+
+        // Make sure enough time has passed from last rebalance to start a new proposal
+        require(block.timestamp >= lastRebalanceTimestamp.add(rebalanceInterval));
+
+        // Set auction parameters
+        rebalancingSet = _rebalancingSet;
```
felix2feng: We should have a check to ensure that the address proposed is indeed a Set, perhaps querying Core

contracts/core/RebalancingSet.sol
felix2feng: Ah I see from below that share Ratio is just the internal naming

contracts/core/RebalancingSet.sol
```diff
@@ -0,0 +1,389 @@
...
+
+    /*
+     * Get addresses of setToken underlying the Rebalancing Set
+     *
+     * @return  componentAddresses       Array of currentSet
+     */
+    function getComponents()
+        external
+        view
+        returns(address[1])
```
felix2feng: Is this more efficient? I think the normal getComponents interface is just address[]. Wonder if there's any issues with address[1]

contracts/core/SetToken.sol
```diff
@@ -286,8 +284,6 @@ contract SetToken is
         public
         returns (bool)
     {
-        // Confirm address is not null
```
felix2feng: Any reason we remove this

contracts/core/RebalancingSet.sol
bweick: RebalancingToken is the verbage we settled on

contracts/core/RebalancingSet.sol
bweick: In my mind this can use either. It's just an address, all someone needs to do is prop up a smart contract that acts as the oracle and set that as the manager.

contracts/core/RebalancingSet.sol
bweick: Cool yeah still not sure what I wanna set that at, but yeah that's a good thing to use.

contracts/core/RebalancingSet.sol
bweick: Yeah I'd be ok with that.

contracts/core/RebalancingSet.sol
bweick: It wouldn't let me return a dynamic address for whatever reason. Perhaps because what's being returned isn't dynamic in nature?

contracts/core/SetToken.sol
bweick: It's duped...that logic already exists in the StandardToken contract which we call through `super`.

contracts/core/RebalancingSet.sol
bweick: Can you explain what you mean? If it's what I think you mean I suppose that would look cleaner but not sure it'd be more efficient.

contracts/core/RebalancingSet.sol
bweick: Sure...I'll leave the organization to @asoong

contracts/core/RebalancingSet.sol
bweick: This might be an issue when we go to issue (lol)...I'll have to consider ways to be able to return a dynamic address array

PR #174, opened 2018-08-16
Add Time-Bound Authorization settings

felix2feng:
- Authorized addresses of the vault, SetTokenFactory, TransferProxy, and TakerWalletWrapper can only be set for a month.
- Unit testing includes traversals and reversions of timestamp

### DIFF
--- a/contracts/core/SetTokenFactory.sol
+++ b/contracts/core/SetTokenFactory.sol
@@ -38,7 +38,9 @@ contract SetTokenFactory
 
     /* ============ Constructor ============ */
     
-    constructor() Authorizable(4 weeks) {}
+    constructor() 
+        Authorizable(4 weeks)
+    {}
 
     /* ============ Setter Functions ============ */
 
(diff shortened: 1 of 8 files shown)


contracts/core/SetTokenFactory.sol
```diff
@@ -36,6 +36,12 @@ contract SetTokenFactory
     // Address of the Core contract
     address public core;
 
+    /* ============ Constructor ============ */
+    
+    constructor() 
+        Authorizable(4 weeks)
```
asoong: How does it calculate 4 weeks? Is it automatically 28 days?

asoong: I dont think SetTokenFactory needs Authorizable, going to change to just be Core's address

felix2feng: It might be more explicit to note do the number of seconds vs. using a higher level construct like weeks.

OK - will remove from SetTokenFactory

felix2feng: @asoong I wonder if it makes sense to pass the Core address in the constructor. Do we imagine needing to change Core's address?

PR #175, opened 2018-08-16
Alex/factory for rebalancing set

asoong:


.solcover.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   skipFiles: [
     'Migrations.sol',
     'mocks',
-    'external'
+    'external',
+    'core/RebalancingTokenFactory.sol'
```
asoong: Ignoring this test suite for now for now. It needs to be tested via core since the transaction sender needs to be an account.

contracts/core/RebalancingTokenFactory.sol
```diff
@@ -0,0 +1,140 @@
...
+     * @param  _name           The name of the new RebalancingToken
+     * @param  _symbol         The symbol of the new RebalancingToken
+     * @param  _callData       Byte string containing additional call parameters
+     * @return setToken        The address of the newly created SetToken
+     */
+    function create(
+        address[] _components,
+        uint256[] _units,
+        uint256 _,
+        string _name,
```
felix2feng: We need to converge on an interface for create. I have a PR where the _name and _symbol are inputted as bytes32

contracts/core/RebalancingTokenFactory.sol
```diff
@@ -0,0 +1,140 @@
...
+     * Set core
+     *
+     * @param  _core   The address of deployed core contract
+     */
+    constructor(
+        address _core
+    )
+        public
+    {
+        core = _core;
```
felix2feng: Nice, if we go with passing the core in the constructor,  we may need to update this pattern in SetFactory too?

contracts/core/interfaces/ICore.sol
```diff
@@ -25,6 +25,15 @@ pragma solidity 0.4.24;
  * various extensions and is a light weight way to interact with the contract.
  */
 interface ICore {
+    /*
+     * Get natural unit of Set
```
felix2feng: The javadocs seems should be validSets vs. for natural unit

PR #176, opened 2018-08-16
[WIP] Update core create interface

asoong:


contracts/core/lib/CoreState.sol
```diff
@@ -134,35 +128,6 @@ contract CoreState {
         return state.factories;
     }
 
-    /**
-     * Return boolean indicating if address is valid Set.
-     *
-     * @param  _set           Set address
-     * @return bool           Boolean indicating if valid Set
-     */
-    function validSets(
```
bweick: This needs to be in Core for issue and redeem. How are we supposed to know which factory to query?

contracts/core/SetTokenFactory.sol
```diff
@@ -64,27 +77,93 @@ contract SetTokenFactory
...
+            }
+        }
+    }
+
+    /**
+     * Check mapping for Set validity
+     *
+     * @param  _set   The address of the SetToken to verify
+     */
+    function isSetValid(
```
felix2feng: Why are we checking the validity of the Set here? Can't we just check Core?

contracts/core/SetTokenFactory.sol
```diff
@@ -28,32 +27,46 @@ import { SetToken } from "./SetToken.sol";
...
 
     // Address of the Core contract
     address public core;
 
-    /* ============ Setter Functions ============ */
+    // Array of tracked SetTokens
+    address[] public setTokens;
+
+    // Mapping of tracked SetTokens
+    mapping(address => bool) public validSets;
```
felix2feng: Don't we do this in Core?

contracts/core/lib/CoreState.sol
```diff
@@ -44,12 +44,6 @@ contract CoreState {
         // Array of tracked SetToken factories
         address[] factories;
 
-        // Mapping of tracked SetTokens
-        mapping(address => bool) validSets;
```
felix2feng: Why are we moving the tracked Sets into factories?

PR #177, opened 2018-08-16
Brian/address array utils

bweick:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -101,13 +102,7 @@ contract CoreInternal is
...
 
         // Find and remove factory from factories array
-        for (uint256 i = 0; i < state.factories.length; i++) {
-            if (state.factories[i] == _factory) {
-                state.factories[i] = state.factories[state.factories.length - 1];
-                state.factories.length -= 1;
-                break;
-            }
-        }
+        state.factories = AddressArrayUtils.remove(state.factories, _factory);
```
felix2feng: Instead, maybe we could do `using AddressArrayUtils for address[]` at the top of the file if that makes things cleaner?

PR #178, opened 2018-08-17
Update interfaces of create's functions to use bytes32 vs string for name/symbol

felix2feng:
- First Stab at utilizing Bytes32 for name and symbol
- Includes parsing the bytes32 into strings


felix2feng:
Gas-wise, I think it's about 20k gas per parse of a 5 letter word? It's not great and we should try to optimize, but its a start. And ya, its only the constructor

contracts/lib/Bytes32.sol
```diff
@@ -0,0 +1,47 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+pragma solidity 0.4.24;
+
+
+library Bytes32 {
+    function bytes32ToBytes(bytes32 data)
+        internal
```
bweick: Do these need to be internal? If we can make them external then they won't compile into the contract. We just have to link the library.

PR #179, opened 2018-08-20
Fix getOwnerBalance order params order to ensure consistency

felix2feng:


contracts/core/Vault.sol
```diff
@@ -172,12 +172,13 @@ contract Vault is
...
-     * @param  _owner           The address of the token owner
      * @param  _token           The address of the ERC20 token
+     * @param  _owner           The address of the token owner
      */
     function getOwnerBalance(
-        address _owner,
-        address _token
+        address _token,
+        address _owner
+
```
justinkchen: remove extra line

justinkchen:
wondering if we should also just change the interfaces for `incrementTokenOwner` and `decrementTokenOwner` too maybe? so its consistent within file?

justinkchen:
also don't forget to we'll need to modify the contract generated wrapper in setprotocol.js as well when we integrate with this change. (and we'll need to `yarn deploy:development`)

felix2feng:
Yeah - I think changing the interfaces  for `incrementTokenOwner` and `decreaseTokenOwner` makes sense.

PR #180, opened 2018-08-20
Remove setCoreAddress in favor of passing through in constructor

felix2feng:


contracts/core/RebalancingTokenFactory.sol
```diff
@@ -22,7 +22,7 @@ import { LibBytes } from "../external/0x/LibBytes.sol";
 
 
 /**
- * @title SetTokenFactory
+ * @title Rebalancing SetTokenFactory
```
bweick: "RebalancingTokenFactory" seems better

PR #181, opened 2018-08-21
[WIP] Brian/auction set up logic

bweick:


contracts/core/RebalancingSetToken.sol
```diff
@@ -376,4 +389,67 @@ contract RebalancingSetToken is
...
+
+        // Get naturalUnit of both sets
+        uint256 currentSetNaturalUnit = currentSetInterface.naturalUnit();
+        uint256 rebalancingSetNaturalUnit = rebalancingSetInterface.naturalUnit();
+
+        // Get units arrays for both sets
+        uint256[] memory currentSetUnits = currentSetInterface.getUnits();
+        uint256[] memory rebalancingSetUnits = rebalancingSetInterface.getUnits();
+
+        for (uint16 i=0; i < combinedTokenArray.length; i++) {
```
felix2feng: Let's use `uint256` vs uint16 - since there are no gas savings

contracts/core/RebalancingSetToken.sol
```diff
@@ -376,4 +389,67 @@ contract RebalancingSetToken is
...
+        }
+    }
+
+    /**
+     * Function to calculate the transfer value of a component given quantity of Set
+     *
+     * @param _unit             The units of the component token
+     * @param _naturalUnit      The natural unit of the Set token
+     */
+    function computeUnits(
```
felix2feng: Do we have one of these functions elsewhere? If so, maybe we can avoid duplication by using a lib

test/core/rebalancingSetToken.spec.ts
```diff
@@ -28,6 +35,10 @@ BigNumberSetup.configure();
 ChaiSetup.configure();
 const { expect } = chai;
 const RebalancingSetToken = artifacts.require('RebalancingSetToken');
+const Core = artifacts.require('CoreMock');
+
+// import { injectInTruffle } from 'sol-trace-set';
```
felix2feng: Remove if not needed?

test/core/rebalancingSetToken.spec.ts
```diff
@@ -28,6 +35,10 @@ BigNumberSetup.configure();
 ChaiSetup.configure();
 const { expect } = chai;
 const RebalancingSetToken = artifacts.require('RebalancingSetToken');
+const Core = artifacts.require('CoreMock');
```
felix2feng: Make naming consistent? `Core` and `CoreMock`. Maybe go with `CoreMock` so that future we don't trip over this.

test/core/rebalancingSetToken.spec.ts
```diff
@@ -41,22 +52,40 @@ contract('RebalancingSetToken', accounts => {
 
   let rebalancingSetToken: RebalancingSetTokenContract;
   let components: StandardTokenMockContract[] = [];
+
+  let core: CoreMockContract;
```
felix2feng: wonder if we should rename this all to `coreMock` to be clear

test/core/rebalancingSetToken.spec.ts
```diff
@@ -41,22 +52,40 @@ contract('RebalancingSetToken', accounts => {
...
   let factory: SetTokenFactoryContract;
+  let rebalancingFactory: RebalancingSetTokenFactoryContract;
 
   const coreWrapper = new CoreWrapper(deployerAccount, deployerAccount);
   const erc20Wrapper = new ERC20Wrapper(deployerAccount);
   const blockchain = new Blockchain(web3);
 
   before(async () => {
     await blockchain.saveSnapshotAsync();
+    ABIDecoder.addABI(Core.abi);
```
felix2feng: `CoreMock`?

test/core/rebalancingSetToken.spec.ts
```diff
@@ -345,18 +466,18 @@ contract('RebalancingSetToken', accounts => {
         const caller = managerAccount;
         const timeFastForward = 100000;
 
-        blockchain.increaseTimeAsync(timeFastForward);
-        await rebalancingSetToken.propose.sendTransactionAsync(
-          newRebalancingToken,
+        await blockchain.increaseTimeAsync(timeFastForward);
```
felix2feng: 😢

test/core/rebalancingSetToken.spec.ts
```diff
@@ -335,6 +377,85 @@ contract('RebalancingSetToken', accounts => {
         await expectRevertError(subject());
       });
     });
+  });
+
+  describe('#mint called from Core', async () => {
+    let rebalancingToken: RebalancingSetTokenContract;
```
felix2feng: As we discussed - a lot of this setup could be moved into a wrapper.

felix2feng:
@asoong Could you help do a review as well?

contracts/core/RebalancingSetToken.sol
bweick: We have something similar in Core but don't see a need to have us make a call to it in order to calculate this.

test/core/rebalancingSetToken.spec.ts
bweick: I don't see a reason to. We're only using CoreMock in this test suite...but it's still what we're using as Core.

test/core/rebalancingSetToken.spec.ts
bweick: si

test/core/rebalancingSetToken.spec.ts
bweick: Yep another PR methinks

asoong: Yes and no, a lot of this stuff is unnecessary such as declaring a new `rebalancingToken` and `setName, setSymbol` (defined in coreWrapper already) that I noticed. I cleaned this up in my PR that should be on master. Makes following along with specs easier

test/core/rebalancingSetToken.spec.ts
asoong: +1, let's use coreMock here when it's coreMock

contracts/core/RebalancingSetToken.sol
```diff
@@ -376,4 +390,67 @@ contract RebalancingSetToken is
...
+            (uint256 indexCurrent, bool isInCurrent) = oldComponents.indexOf(combinedTokenArray[i]);
+            (uint256 indexRebalance, bool isInRebalance) = newComponents.indexOf(combinedTokenArray[i]);
+
+            // Compute and push unit amounts of token in currentSet, push 0 if not in set
+            if (isInCurrent) {
+                combinedCurrentUnits.push(
+                    computeUnits(currentSetUnits[indexCurrent], currentSetNaturalUnit)
+                );
+            } else {
+                combinedCurrentUnits.push(uint256(0));
```
asoong: interesting, do you have to cast this here? `uint256(0)`?

test/core/rebalancingSetToken.spec.ts
```diff
@@ -863,9 +1173,9 @@ contract('RebalancingSetToken', accounts => {
         const caller = managerAccount;
         const timeFastForward = 100000;
 
-        blockchain.increaseTimeAsync(timeFastForward);
+        await blockchain.increaseTimeAsync(timeFastForward);
```
asoong: Would really love for this whole beforeEach to be one wrapper function that encompasses all the state transitions

test/core/rebalancingSetToken.spec.ts
```diff
@@ -915,34 +1225,61 @@ contract('RebalancingSetToken', accounts => {
...
+    let newRebalancingSetToken: SetTokenContract;
+
+    const naturalUnit: BigNumber = ether(2);
 
     beforeEach(async () => {
-      components = await erc20Wrapper.deployTokensAsync(2, deployerAccount);
+      components = await erc20Wrapper.deployTokensAsync(3, deployerAccount);
+      await erc20Wrapper.approveTransfersAsync(components, transferProxy.address);
+
+      const currentComponentAddresses = _.map(components.slice(0, 2), token => token.address);
```
asoong: Define a constant for this slice

test/core/rebalancingSetToken.spec.ts
```diff
@@ -915,34 +1225,61 @@ contract('RebalancingSetToken', accounts => {
...
+      const currentComponentUnits = _.map(components.slice(0, 2), () => naturalUnit.mul(2)); // Multiple of naturalUnit
+      currentSetToken = await coreWrapper.createSetTokenAsync(
+        core,
+        factory.address,
+        currentComponentAddresses,
+        currentComponentUnits,
+        naturalUnit,
+      );
+
+      const newComponentAddresses = _.map(components.slice(1, 3), token => token.address);
```
asoong: Same

test/core/rebalancingSetToken.spec.ts
```diff
@@ -252,34 +279,47 @@ contract('RebalancingSetToken', accounts => {
...
+    let currentSetToken: SetTokenContract;
+
+    const naturalUnit: BigNumber = ether(2);
 
     beforeEach(async () => {
       components = await erc20Wrapper.deployTokensAsync(2, deployerAccount);
-      factory = await coreWrapper.deploySetTokenFactoryAsync(coreAccount);
+      await erc20Wrapper.approveTransfersAsync(components, transferProxy.address);
+
+      const currentComponentAddresses = _.map(components.slice(0, 2), token => token.address);
```
asoong: Why do you need to slice here?

asoong:
Doing some cleanup now will help you in the next phase. @felix2feng  In the future, let's keep the PR smaller. Maybe start with `CoreMock#mint` specs first before moving on to burn.

contracts/core/RebalancingSetToken.sol
bweick: Prob not but feel like we've fallen on the standard of explicitly stating what all the uints are

test/core/rebalancingSetToken.spec.ts
```diff
@@ -252,34 +279,47 @@ contract('RebalancingSetToken', accounts => {
     });
   });
 
-  describe('#mint', async () => {
+  describe('#mint called directly', async () => {
```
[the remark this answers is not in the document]
bweick: Agree mostly do you think it makes sense to keep one "happy" path case in the "direct" case so that we aren't accidentally missing something there?

utils/RebalancingTokenWrapper.ts
```diff
@@ -0,0 +1,166 @@
...
+      core,
+      factory,
+      [initialSet],
+      [initialUnitShares],
+      DEFAULT_REBALANCING_NATURAL_UNIT,
+      callData,
+    );
+  }
+
+  public async defaultTransitionToPropose(
```
asoong: Why are these prefixed to default?

utils/RebalancingTokenWrapper.ts
```diff
@@ -0,0 +1,166 @@
...
+    blockchain: Blockchain,
+  ) {
+    this._tokenOwnerAddress = tokenOwnerAddress;
+
+    this._coreWrapper = coreWrapper;
+    this._erc20Wrapper = erc20Wrapper;
+    this._blockchain = blockchain;
+  }
+
+  public async createSetTokens(
```
asoong: suffix methods `Async`

utils/RebalancingTokenWrapper.ts
```diff
@@ -0,0 +1,166 @@
...
+    core: CoreLikeContract,
+    factory: Address,
+    transferProxy: Address,
+    from: Address = this._tokenOwnerAddress,
+  ): Promise<SetTokenContract[]> {
+    const naturalUnit = ether(2);
+    const components = await this._erc20Wrapper.deployTokensAsync(3, this._tokenOwnerAddress);
+    await this._erc20Wrapper.approveTransfersAsync(components, transferProxy);
+
+    const set1Components = components.slice(0, 2);
```
asoong: nit: don't use numbers. `firstSetComponents`

utils/constants.ts
```diff
@@ -14,6 +14,13 @@ export const UNLIMITED_ALLOWANCE_IN_BASE_UNITS = new BigNumber(2).pow(256).minus
 export const MAX_DIGITS_IN_UNSIGNED_256_INT = 78;
 export const ZRX_TOKEN_ADDRESS = '0x871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c';
 
+// Rebalancing Constants
+export const DEFAULT_PERIOD_INTERVAL = new BigNumber(90000);
```
asoong: Helps to be even more realistic here for clarity. Your contracts specify day cutoffs. Create constants for maybe `ONE_DAY_IN_SECONDS` and your test case for invalid `PERIOD_INTERVAL` can be `ONE_DAY_IN_SECONDS.sub(1)`

asoong: It makes your usage more readable: `const proposalPeriod = ONE_DAY_IN_SECONDS;` vs.
`const proposalPeriod = DEFAULT_PERIOD_INTERVAL;`

utils/RebalancingTokenWrapper.ts
felix2feng: This doesn't seem rebalancing specific. Should this be in CoreWrapper instead?

utils/RebalancingTokenWrapper.ts
```diff
@@ -0,0 +1,166 @@
...
+      caller
+    );
+
+    await this._blockchain.increaseTimeAsync(DEFAULT_TIME_FAST_FORWARD);
+    await rebalancingSetToken.rebalance.sendTransactionAsync(
+      { from: caller, gas: DEFAULT_GAS }
+    );
+  }
+
+  public async constructCombinedUnitArray(
```
felix2feng: Could we add a javadoc on what this does?

test/core/rebalancingSetToken.spec.ts
```diff
@@ -302,6 +352,63 @@ contract('RebalancingSetToken', accounts => {
...
+  describe('#mint: Called from CoreMock', async () => {
+    let rebalancingSetToken: RebalancingSetTokenContract;
+    let subjectIssuer: Address;
+    let subjectQuantity: BigNumber;
+    let subjectCaller: Address;
+
+    let newRebalancingSetToken: SetTokenContract;
+
+    beforeEach(async () => {
+      const setTokens = await rebalancingTokenWrapper.createSetTokens(coreMock, factory.address, transferProxy.address);
```
felix2feng: You could combine this and the next two lines in one line
```
const [ currentSetToken, newRebalancingSetToken ] = await rebalancingTokenWrapper.createSetTokens...
```

test/core/rebalancingSetToken.spec.ts
```diff
@@ -302,6 +352,63 @@ contract('RebalancingSetToken', accounts => {
...
+      const proposalPeriod = DEFAULT_PERIOD_INTERVAL;
+      rebalancingSetToken = await rebalancingTokenWrapper.createDefaultRebalancingSetToken(
+        coreMock,
+        rebalancingFactory.address,
+        managerAccount,
+        currentSetToken.address,
+        proposalPeriod
+      );
+
+
```
felix2feng: extra new line

test/core/rebalancingSetToken.spec.ts
```diff
@@ -367,48 +449,121 @@ contract('RebalancingSetToken', accounts => {
...
 
-    let newRebalancingToken: Address;
-
     beforeEach(async () => {
+      const naturalUnit: BigNumber = ether(2);
       components = await erc20Wrapper.deployTokensAsync(2, deployerAccount);
-      factory = await coreWrapper.deploySetTokenFactoryAsync(coreAccount);
+      await erc20Wrapper.approveTransfersAsync(components, transferProxy.address);
+
+      const currentComponentAddresses = _.map(components, token => token.address);
```
felix2feng: Could we put the map function in a util or something? It's a bit jarring to see this low level stuff in a test

bweick: Those tests are gonna be deleted anyway but I also don't think it's worthwhile to create a full other util for one line of code.

bweick: Sorry not gonna be deleted was looking at wrong test suite but still don't think its worth creating its own util... _.map is the util.

utils/RebalancingTokenWrapper.ts
bweick: If we wanna move it at some point we can. Just feels like the only functionality we need multiple SetTokens for rn is rebalancing.

utils/RebalancingTokenWrapper.ts
bweick: default -> propose state

test/core/rebalancingSetToken.spec.ts
bweick: Different scoping on them though

PR #188, opened 2018-08-22
Use shared methods from set protocol utils

asoong:
You can merge your stuff in @bweick and i'll work around you. Just make sure to rebase for your next PR.


test/core/extensions/coreIssuance.spec.ts
```diff
@@ -162,7 +162,7 @@ contract('CoreIssuance', accounts => {
 
     describe('when the set was not created through core', async () => {
       beforeEach(async () => {
-        subjectSetToIssue = NULL_ADDRESS;
+        subjectSetToIssue = SetProtocolUtils.CONSTANTS.NULL_ADDRESS;
```
felix2feng: Maybe we can declare
```
const { NULL_ADDRESS }` =  SetProtocolUtils.CONSTANTS;
```
above, so we only need to reference NULL_ADDRESS in the tests themselves

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -317,9 +317,9 @@ contract('CoreIssuanceOrder', accounts => {
 
         // Standard 0x order without fees, see zeroExExchangeWrapper.spec.ts for clarity on body
         const zeroExOrder: ZeroExOrder = Utils.generateZeroExOrder(
-          NULL_ADDRESS,                       // senderAddress
+          Utils.CONSTANTS.NULL_ADDRESS,       // senderAddress
```
felix2feng: Same here re: declaration on top

test/core/extensions/coreIssuance.spec.ts
asoong: +1

test/core/extensions/coreIssuanceOrder.spec.ts
asoong: +1

PR #193, opened 2018-08-24
Brian/rebalancing token misc

bweick:


contracts/core/RebalancingSetToken.sol
```diff
@@ -171,21 +171,22 @@ contract RebalancingSetToken is
         require(msg.sender == manager);
 
         // New proposal cannot be made during a rebalance period
-        require(rebalanceState == State.Default);
+        require(rebalanceState != State.Rebalance);
```
asoong: Is this so they can re-propose?

contracts/core/RebalancingSetToken.sol
```diff
@@ -171,21 +171,22 @@ contract RebalancingSetToken is
...
 
         // Make sure enough time has passed from last rebalance to start a new proposal
         require(block.timestamp >= lastRebalanceTimestamp.add(rebalanceInterval));
 
+        // Check that new proposed Set is valid Set created by Core
+        address core = ISetFactory(factory).core();
```
asoong: `core` doesn't look re-used. Insert `ISetFactory(factory).core()` directly into the next expression

contracts/core/RebalancingSetToken.sol
```diff
@@ -327,9 +331,11 @@ contract RebalancingSetToken is
     function getComponents()
         external
         view
-        returns(address[1])
+        returns(address[])
     {
-        return [currentSet];
+        address[] memory components = new address[](1);
```
asoong: this seems verbose. with only one value that is already known, you can just initialize the array with it I think: https://ethereum.stackexchange.com/questions/28047/solidity-initialize-memory-array-of-fixed-size

contracts/core/RebalancingSetToken.sol
```diff
@@ -340,9 +346,11 @@ contract RebalancingSetToken is
     function getUnits()
         external
         view
-        returns(uint256[1])
+        returns(uint256[])
     {
-        return [unitShares];
+        uint256[] memory units = new uint256[](1);
```
asoong: same

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -18,7 +21,15 @@ import {
 import { ether } from '../../../utils/units';
 import { IssuanceComponentDeposited } from '../../../utils/contract_logs/core';
 import { assertTokenBalance, expectRevertError } from '../../../utils/tokenAssertions';
-import { DEFAULT_GAS, DEPLOYED_TOKEN_QUANTITY, ZERO } from '../../../utils/constants';
+import {
+  DEFAULT_GAS,
+  DEPLOYED_TOKEN_QUANTITY,
+  NULL_ADDRESS,
```
asoong: How is this working? I think you need to rebase, `NULL_ADDRESS` is part of set-protocol-utils now

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -316,7 +330,242 @@ contract('CoreIssuance', accounts => {
...
+        setTokenFactory.address,
+        componentAddresses,
+        componentUnits,
+        naturalUnit,
+      );
+
+      initialShareRatio = DEFAULT_UNIT_SHARES;
+      rebalancingNaturalUnit = DEFAULT_REBALANCING_NATURAL_UNIT;
+      const initialSet = setToken.address;
+      const manager = managerAccount;
```
asoong: Can you move this into a utils method or something? Something like:
`generateRebalancingSetCreatorCallData` and you can move all of these one time use constants into there.

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -316,7 +330,242 @@ contract('CoreIssuance', accounts => {
...
+          { from: ownerAccount, gas: DEFAULT_GAS },
+        );
+      });
+
+      it('should revert', async () => {
+        await expectRevertError(subject());
+      });
+    });
+
+    // describe('when the quantity is not a multiple of the natural unit of the set', async () => {
```
asoong: ?

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -442,6 +691,148 @@ contract('CoreIssuance', accounts => {
...
+      componentUnits = _.map(components, () => ether(4)); // Multiple of naturalUnit
+      setToken = await coreWrapper.createSetTokenAsync(
+        core,
+        setTokenFactory.address,
+        componentAddresses,
+        componentUnits,
+        naturalUnit,
+      );
+
+      initialShareRatio = DEFAULT_UNIT_SHARES;
```
asoong: same as above for this setup

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -316,7 +330,242 @@ contract('CoreIssuance', accounts => {
...
+      const initialSet = setToken.address;
+      const manager = managerAccount;
+      const proposalPeriod = ONE_DAY_IN_SECONDS;
+      const rebalanceInterval = ONE_DAY_IN_SECONDS;
+      const callData = Utils.bufferArrayToHex([
+        Utils.paddedBufferForPrimitive(manager),
+        Utils.paddedBufferForBigNumber(proposalPeriod),
+        Utils.paddedBufferForBigNumber(rebalanceInterval),
+      ]);
+      rebalancingToken = await coreWrapper.createRebalancingTokenAsync(
```
asoong: for this `createRebalancingTokenAsync` method, make some of the parameters default to these constants so you dont need to keep defining them above

test/core/extensions/coreIssuance.spec.ts
```diff
@@ -442,6 +691,148 @@ contract('CoreIssuance', accounts => {
...
+      beforeEach(async () => {
+        subjectQuantityToRedeem = ether(3);
+      });
+
+      it('should revert', async () => {
+        await expectRevertError(subject());
+      });
+    });
+
+    // describe('when the quantity is not a multiple of the natural unit of the set', async () => {
```
asoong: ?

test/core/setToken.spec.ts
```diff
@@ -390,145 +390,4 @@ contract('SetToken', accounts => {
       });
     });
   });
-
-  describe('#transfer', async () => {
```
asoong: why are you removing these?

contracts/core/RebalancingSetToken.sol
bweick: Si papi

contracts/core/RebalancingSetToken.sol
asoong: try `return new address[currentSet];`

contracts/core/RebalancingSetToken.sol
bweick: Yeah felt like that would be super messy though. Preferred the clarity.

test/core/setToken.spec.ts
bweick: It's unnecessary...it basically calls the inherited xfer fxn except has a check to make sure that the recipient address isn't the smart contract itself. Just no need to have it in my mind.

contracts/core/RebalancingSetToken.sol
asoong: 👍

contracts/core/RebalancingSetToken.sol
asoong: going for gas optimization at the contract level, extra memory consumed here

contracts/core/RebalancingSetToken.sol
bweick: The problem is that, that doesn't return a dynamic typed array which wouldn't conform to the getComponents and getUnits interface

PR #194, opened 2018-08-25
Add generated types exports

justinkchen:
Update artifacts, add generated types from abi-gen to package exports, fix up tsconfig, fixed the contract generator template


tsconfig.json
```diff
@@ -2,14 +2,18 @@
   "compilerOptions": {
     "outDir": "./transpiled/",
     "sourceMap": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
felix2feng: Has this been an issue with any dependencies?

utils/Web3Utils.ts
```diff
@@ -0,0 +1,119 @@
+/*
```
felix2feng: This seems to be a good candidate to put in the utils repo? where did this come from?

types/base_contract.ts
```diff
@@ -35,7 +42,7 @@ export class BaseContract {
         };
         if (_.isUndefined(txDataWithDefaults.gas) && !_.isUndefined(estimateGasAsync)) {
             const estimatedGas = await estimateGasAsync(txData);
-            txDataWithDefaults.gas = estimatedGas;
+            txDataWithDefaults.gas = new BigNumber(estimatedGas);
```
felix2feng: What was the need for the change here?

felix2feng:
😢 Coverage. Error is:
```
Error: Cannot find module '../../artifacts/ts/Authorizable'
```

types/base_contract.ts
justinkchen: this is just what was being done in the setprotocol.js side so decided to do it here

utils/Web3Utils.ts
justinkchen: we'be been using it in both sides but its based off of the web3-utils package

tsconfig.json
justinkchen: causes a lot of issues with the artifacts that are autogenerated

PR #197, opened 2018-08-28
Brian/bidding

bweick:


contracts/core/RebalancingSetToken.sol
```diff
@@ -245,6 +248,81 @@ contract RebalancingSetToken is
         rebalanceState = State.Default;
     }
 
+    /*
+     * Initiate settlement for the rebalancing set. Full functionality now returned to
```
justinkchen: Wrong comment

contracts/core/interfaces/IVault.sol
```diff
@@ -73,20 +73,36 @@ interface IVault {
...
         address _token,
         uint256 _quantity
     )
         external;
 
+    /**
+     * Transfers tokens associated with one account to another account in the vault
+     *
+     * @param  _to               Address tokens being transferred to
+     * @param  _from             Address tokens being transferred from
```
justinkchen: switch order of params in comment

contracts/core/RebalancingSetToken.sol
```diff
@@ -245,6 +248,81 @@ contract RebalancingSetToken is
...
+        remainingCurrentSets = remainingCurrentSets.sub(filled_quantity);
+        rebalanceSetSupply = rebalanceSetSupply.add(rebalanceSetsAdded);
+        return (combinedTokenArray, inflowUnitArray, outflowUnitArray);
+    }
+
+    function getBidPrice(
+        uint256 _quantity
+    )
+        public
+        returns (uint256[], uint256[], uint256)
```
asoong: pure?

contracts/core/Vault.sol
```diff
@@ -186,4 +210,24 @@ contract Vault is
         // Return owners token balance
         return balances[_token][_owner];
     }
+
+    /* ============ Internal Functions ============ */
+
+    function transferBalanceInternal(
```
asoong: The naming scheme in the vault file is a little weird. A few suggestions:
- Does it make sense for it to be caleld `incrementTokenOwner` and `decrementTokenOwner`? Everything else is suffixed with `Balance` now which I like since it's explicit about the property being mutated
- CoreAccounting has an `internalTransfer`. Let's rename this to something like `transferVaultBalanceInternal`, keeping to the `internal` suffix and being explicit that it's for vault only

contracts/core/RebalancingSetToken.sol
```diff
@@ -245,6 +248,81 @@ contract RebalancingSetToken is
...
+    {
+        // Confirm in Rebalance State
+        require(rebalanceState == State.Rebalance);
+
+        // Declare unit arrays in memory
+        uint256[] memory inflowUnitArray = new uint256[](combinedTokenArray.length);
+        uint256[] memory outflowUnitArray = new uint256[](combinedTokenArray.length);
+
+        // Get bid conversion price
+        uint256 priceNumerator = 1;
```
asoong: Why are these variables? It doesn't seem like they get updated

contracts/core/RebalancingSetToken.sol
```diff
@@ -245,6 +248,81 @@ contract RebalancingSetToken is
...
+        uint256 priceDivisor = 1;
+
+        for (uint256 i=0; i < combinedTokenArray.length; i++) {
+            uint256 rebalanceUnit = combinedRebalanceUnits[i];
+            uint256 currentUnit = combinedCurrentUnits[i];
+
+            // If rebalance greater than currentUnit*price token inflow, else token outflow
+            if (rebalanceUnit > currentUnit.mul(priceNumerator).div(priceDivisor)) {
```
asoong: `.mul(priceNumerator).div(priceDivisor)` do this math outside of the for loop

contracts/core/RebalancingSetToken.sol
```diff
@@ -245,6 +248,81 @@ contract RebalancingSetToken is
...
+                inflowUnitArray[i] = _quantity.mul(rebalanceUnit.sub(
+                    priceNumerator.mul(currentUnit).div(priceDivisor)
+                )).div(10**18);
```
asoong: Declare this outside for loop for reuse

PR #207, opened 2018-09-18
Add natural unit restrictions for proposed Set. Create minimumBid amo…

bweick:
…unt. Redo token flow math to use max component Set Token natural unit as base amount instead of 10**18. Figure out how many Sets to issue in settlement and recalcualte unitShares.


contracts/core/RebalancingSetToken.sol
```diff
@@ -16,14 +16,18 @@
 
 pragma solidity 0.4.24;
 
+import { AddressArrayUtils } from "../external/cryptofin/AddressArrayUtils.sol";
```
felix2feng: FYI - If you rebase with master, this import now uses the CryptoFin import.

contracts/core/RebalancingSetToken.sol
```diff
@@ -47,7 +51,7 @@ contract RebalancingSetToken is
     /* ============ State Variables ============ */
 
     address public factory;
-    uint256 public naturalUnit = 1;
+    uint256 public naturalUnit = 10**10;
```
felix2feng: Add a comment explaining why the natural unit is 10 ** 10.

contracts/core/RebalancingSetToken.sol
```diff
@@ -151,14 +156,14 @@ contract RebalancingSetToken is
      * Function used to set the terms of the next rebalance and start the proposal period
      *
      *
-     * @param _rebalancingSet               The Set to rebalance into
+     * @param _nextSet               The Set to rebalance into
```
felix2feng: Knit: fix comment spacing

contracts/core/RebalancingSetToken.sol
```diff
@@ -84,7 +89,7 @@ contract RebalancingSetToken is
     );
 
     event RebalanceProposed(
-        address rebalancingSet,
+        address nextSet,
```
felix2feng: 💯 nextSet is a way better name

contracts/core/RebalancingSetToken.sol
```diff
@@ -311,29 +354,29 @@ contract RebalancingSetToken is
...
-        uint256 priceNumerator = 1;
-        uint256 priceDivisor = 1;
+        uint256 priceNumerator = 1374;
+
+        // Normalized quantity amount
+        uint256 unitsMultiplier = _quantity.div(minimumBid).mul(auctionPriceDivisor);
 
         for (uint256 i=0; i < combinedTokenArray.length; i++) {
-            uint256 rebalanceUnit = combinedRebalanceUnits[i];
+            uint256 nextUnit = combinedNextUnits[i];
```
felix2feng: Upon first glance, `nextUnit` seems like a confusing name. To be more discrete, maybe use something like `nextSetUnit`.

contracts/core/RebalancingSetToken.sol
```diff
@@ -486,38 +529,48 @@ contract RebalancingSetToken is
...
 
     /*
-     * Get combinedRebalanceUnits of Rebalancing Set
+     * Get combinedNextUnits of Rebalancing Set
      *
-     * @return  combinedRebalanceUnits
+     * @return  combinedNextUnits
      */
-    function getCombinedRebalanceUnits()
+    function getCombinedNextUnits()
```
felix2feng: Maybe `getCombinedNextSetUnits`

contracts/core/RebalancingSetToken.sol
```diff
@@ -533,17 +586,62 @@ contract RebalancingSetToken is
...
         }
     }
 
+    /**
+     * Calculate the amount of nextSets to issue by using the component amounts in the
+     * vault, unitShares following from this calculation.
+     * @return uint256              Amount of nextSets to issue
+     * @return uint256              New unitShares for the rebalancingSetToken
+     */
+    function getIssueAmount()
```
felix2feng: We can name this better. It is currently too general and doesn't convey the work done. Perhaps something like `calculateNextSetIssueQuantity`.

contracts/core/RebalancingSetToken.sol
```diff
@@ -178,10 +185,18 @@ contract RebalancingSetToken is
...
 
         // Check that new proposed Set is valid Set created by Core
-        require(ICore(ISetFactory(factory).core()).validSets(_rebalancingSet));
+        require(ICore(ISetFactory(factory).core()).validSets(_nextSet));
+
+        // Check that the propoosed set is a multiple of current set, or vice versa
+        uint256 currentNaturalUnit = ISetToken(currentSet).naturalUnit();
+        uint256 rebalancingNaturalUnit = ISetToken(_nextSet).naturalUnit();
+        require(
+            Math.max256(currentNaturalUnit, rebalancingNaturalUnit) %
```
felix2feng: If we're going with `nextSet`, then the name of `rebalancingNaturalUnit` should be `nextSetNaturalUnit` or `nextNaturalUnit`

contracts/core/RebalancingSetToken.sol
```diff
@@ -212,13 +227,13 @@ contract RebalancingSetToken is
...
         // Create token arrays needed for auction
-        parseUnitArrays();
+        auctionSetUp();
 
         // Get core address
         address core = ISetFactory(factory).core();
 
         // Calculate remainingCurrentSets
-        remainingCurrentSets = unitShares.mul(totalSupply_);
+        remainingCurrentSets = unitShares.mul(totalSupply_).div(naturalUnit);
```
felix2feng: It seems that making the calculation a function would makes things clearer and less error-prone. 

Maybe we have a function for this calculation already?

contracts/core/RebalancingSetToken.sol
```diff
@@ -241,8 +256,36 @@ contract RebalancingSetToken is
         // Must be in Rebalance state to call settlement
         require(rebalanceState == State.Rebalance);
 
+        // Make sure all currentSets have been rebalanced
```
felix2feng: This comment refers to the name of the function.

The method/function name `settlement` seems more like a process than a verb. Can we rename the function to `settleRebalance` or `resolveReblance` or something better?

contracts/core/RebalancingSetToken.sol
```diff
@@ -241,8 +256,36 @@ contract RebalancingSetToken is
...
+        require(remainingCurrentSets == 0);
+
+        // Create ICore object
```
felix2feng: The code is clear enough that we don't need a comment.

contracts/core/RebalancingSetToken.sol
```diff
@@ -241,8 +256,36 @@ contract RebalancingSetToken is
...
+
+        // Issue nextSet and deposit in Vault
+        uint256 issueAmount;
+        (issueAmount, unitShares) = calculateNextSetIssueQuantity();
+        core.issue(
+            nextSet,
+            issueAmount
+        );
+
+        // Ensure allowance to transfer sets to Vault
```
felix2feng: Don't need comment.

contracts/core/RebalancingSetToken.sol
```diff
@@ -241,8 +256,36 @@ contract RebalancingSetToken is
         // Must be in Rebalance state to call settlement
         require(rebalanceState == State.Rebalance);
 
+        // Make sure all currentSets have been rebalanced
+        require(remainingCurrentSets == 0);
+
+        // Create ICore object
+        ICore core = ICore(ISetFactory(factory).core());
+
+        // Issue nextSet and deposit in Vault
```
felix2feng: This step doesn't deposit into the vault does it?

contracts/core/RebalancingSetToken.sol
```diff
@@ -312,29 +357,29 @@ contract RebalancingSetToken is
         uint256[] memory outflowUnitArray = new uint256[](combinedTokenArray.length);
 
         // Get bid conversion price
-        uint256 priceNumerator = 1;
-        uint256 priceDivisor = 1;
+        uint256 priceNumerator = 1374;
```
felix2feng: What is `1374`? Seems like a magic number.

Instead, make it a constant on the top of the file and ascribe a name to it.

contracts/core/RebalancingSetToken.sol
```diff
@@ -312,29 +357,29 @@ contract RebalancingSetToken is
...
+            uint256 nextUnit = combinedNextSetUnits[i];
             uint256 currentUnit = combinedCurrentUnits[i];
 
             // If rebalance greater than currentUnit*price token inflow, else token outflow
-            if (rebalanceUnit > currentUnit.mul(priceNumerator).div(priceDivisor)) {
-                inflowUnitArray[i] = _quantity.mul(rebalanceUnit.sub(
-                    priceNumerator.mul(currentUnit).div(priceDivisor)
-                )).div(10**18);
+            if (nextUnit > currentUnit.mul(priceNumerator).div(auctionPriceDivisor)) {
+                inflowUnitArray[i] = unitsMultiplier.mul(
```
felix2feng: This seems really convoluted. Can we create a method for these calculations?

contracts/core/RebalancingSetToken.sol
bweick: Ah yes I agree...missed that one

contracts/core/RebalancingSetToken.sol
bweick: Don't see the reason to be removing comments. They're free in my mind.

contracts/core/RebalancingSetToken.sol
bweick: +1

contracts/core/RebalancingSetToken.sol
bweick: It's more a placeholder for when we put in an auction library. It will not be there in the final iteration.

contracts/core/RebalancingSetToken.sol
bweick: I guess what do you hope to gain by making it a function? It doesn't change the logic, it's pretty much already in its own function, I don't see a need to make another one.

contracts/core/RebalancingSetToken.sol
bweick: Sure I can change...don't feel strongly either way

contracts/core/RebalancingSetToken.sol
bweick: I suppose we do on our other contracts but not on this one. Feels like over doing it to make it a function on this one where it wouldn't be used elsewhere. This wasn't an error, we just used to have the naturalUnit be 1 so there was no need to do the division, but with changing it to 10**10 it now needed to be included in the calculation.

contracts/core/RebalancingSetToken.sol
felix2feng: Yeah, the other functions are verbs. Let's rename for consistency.

contracts/core/RebalancingSetToken.sol
felix2feng: If the code is clear enough, we don't need explanatory comments (which is a good thing). I'm not opinionated about this one. So up to you.

contracts/core/RebalancingSetToken.sol
felix2feng: Ok, please leave a comment that this is a placeholder

contracts/core/RebalancingSetToken.sol
felix2feng: An additional function allows us to express the intent of this 5-step arithmetic expression.

IMO, it's really unclear what it is supposed to do. Thus, its harder to audit and ensure the security of it as well.

bweick: Hmm I'm not sure there's a way to clearly get across exactly what this is supposed to do unless we have a derivation of this formula in the code base. I don't think adding a function will enforce anymore clarity cause they'll still look at this formula and wonder how it was derived. Perhaps an insightful multi-line comment is a better way to go about it?

felix2feng: I feel a multi-line doesn't do it justice. you'd have to explain it everywhere you put it.

It seems like the same formula is used twice. With a function, we'd at least enforce that both operate similarly.

The function name itself can be a hint for what the calculations are supposed to be. And we can have a more in-depth function comment in that function too?

bweick: It's only used here, and I guess I look at the whole thing as one formula. it feels like you're saying let's abstract all five-step functions that have these order of operations. What's important is understanding how token flows are calculated, another function in this case, does not solve that problem I don't beleive.

contracts/core/RebalancingSetToken.sol
asoong: I think above you talked about maintaining consistency and a long time ago we agreed on comments on every line and not assuming how obvious it is.

That being said, the comments could have a bit more color. In the case above, it could be, "Creating pointer to Core to Issue next set and Deposit into vault". This one could be: "Ensure transfer proxy has enough spender allowance to move issued nextSet to vault"

contracts/core/RebalancingSetToken.sol
asoong: Omitting constants unless part of a library: https://ethereum.stackexchange.com/questions/13851/could-we-call-a-constant-function-without-spending-any-gas-inside-a-transaction/13855

Leave a comment for declarations like these

contracts/core/RebalancingSetToken.sol
asoong: +1, add a private method for this unit array calculation, and you can abstract the formula explanation into that private method. The comments could be improved/some lines are missing some.

contracts/core/RebalancingSetToken.sol
```diff
@@ -487,38 +527,48 @@ contract RebalancingSetToken is
...
+     * Calcualate unit difference between both sets relative to the largest natural
+     * unit of the two sets.
+     */
+    function auctionSetUp()
         internal
     {
         // Create interfaces for interacting with sets
         ISetToken currentSetInterface = ISetToken(currentSet);
-        ISetToken rebalancingSetInterface = ISetToken(rebalancingSet);
+        ISetToken nextSetInterface = ISetToken(nextSet);
```
asoong: Omit this interface suffix

contracts/core/RebalancingSetToken.sol
```diff
@@ -534,15 +584,61 @@ contract RebalancingSetToken is
...
+        IVault vault = IVault(vaultAddress);
+
+        for (uint256 i = 0; i < nextComponents.length; i++) {
+            // Get amount of components in vault owned by rebalancingSetToken
+            uint256 componentAmount = vault.getOwnerBalance(
+                nextComponents[i],
+                this
+            );
+
+            // Calculate amount of Sets that can be issued from those components, if less than amount for other
```
asoong: Examples of great comments

contracts/core/RebalancingSetToken.sol
```diff
@@ -534,15 +584,61 @@ contract RebalancingSetToken is
...
+            if (componentIssueAmount < maxIssueAmount) {
+                maxIssueAmount = componentIssueAmount;
             }
         }
+
+        // Calculate the amount of naturalUnits worth of rebalancingSetToken outstanding
+        uint256 naturalUnitsOutstanding = totalSupply_.div(naturalUnit);
+
+        // Issue amount of Sets that is closest multiple of nextNaturalUnit to the maxIssueAmount
+        uint256 issueAmount = maxIssueAmount.div(nextNaturalUnit).mul(nextNaturalUnit);
```
asoong: Forgot, where does the dust that is not used to issue remain and who is the owner?

contracts/core/interfaces/ICore.sol
```diff
@@ -25,10 +25,30 @@ pragma solidity 0.4.24;
...
+        public
+        view
+        returns(address);
+
     /*
-     * Get natural unit of Set
+     * Returns if valid set
      *
-     * @return  uint256       Natural unit of Set
+     * @return  bool      If valid set
```
asoong: Improve message here: "Returns whether set was...."

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -253,41 +253,42 @@ contract('CoreIssuanceOrder', accounts => {
 
     it('transfers the full maker token amount from the maker', async () => {
       const existingBalance = await makerToken.balanceOf.callAsync(issuanceOrderMaker);
-      await assertTokenBalance(makerToken, DEPLOYED_TOKEN_QUANTITY, issuanceOrderMaker);
+      await assertTokenBalanceAsync(makerToken, DEPLOYED_TOKEN_QUANTITY, issuanceOrderMaker);
```
asoong: Next time make these changes in a separate PR and rebase yours onto it

asoong: Ignoring spec files unrelated to rebalancing

test/core/extensions/coreRebalanceAuction.spec.ts
```diff
@@ -34,11 +34,13 @@ const { expect } = chai;
 const blockchain = new Blockchain(web3);
 
 
+
```
asoong: ?

contracts/core/RebalancingSetToken.sol
bweick: Remains in the vault under the rebalancingSetToken's "ownership" so could be used in a future rebalance.

utils/RebalancingTokenWrapper.ts
```diff
@@ -47,27 +49,47 @@ export class RebalancingTokenWrapper {
...
     for (const index in indexArray) {
+      let minDec: number;
       const idx = Number(index);
+      const decOne = await components[idx].decimals.callAsync();
+      const decTwo = await components[idx + 1].decimals.callAsync();
+
+      // Determine minimum natural unit if not passed in
+      if (naturalUnits) {
+        naturalUnit = naturalUnits[idx];
+        minDec = 18 - naturalUnit.e;
```
felix2feng: What is .e?

utils/RebalancingTokenWrapper.ts
```diff
@@ -47,27 +49,47 @@ export class RebalancingTokenWrapper {
...
-    const naturalUnit = ether(2);
+    let naturalUnit: BigNumber;
     const setTokenArray: SetTokenContract[] = [];
 
     const components = await this._erc20Wrapper.deployTokensAsync(tokenCount + 1, this._tokenOwnerAddress);
     await this._erc20Wrapper.approveTransfersAsync(components, transferProxy);
 
     const indexArray = _.times(tokenCount, Number);
     for (const index in indexArray) {
+      let minDec: number;
```
felix2feng: Let's favor being verbose and descriptive in variable names. 

`minimumDecimal` vs `minDec`

utils/RebalancingTokenWrapper.ts
```diff
@@ -47,27 +49,47 @@ export class RebalancingTokenWrapper {
...
       const idx = Number(index);
+      const decOne = await components[idx].decimals.callAsync();
```
felix2feng: Same here re: verbose

`componentOneDecimal` vs `decOne`

utils/RebalancingTokenWrapper.ts
```diff
@@ -47,27 +49,47 @@ export class RebalancingTokenWrapper {
...
+      } else {
+        minDec = Math.min(decOne.toNumber(), decTwo.toNumber());
+        naturalUnit = new BigNumber(10 ** (18 - minDec));
+      }
+
+      // Get Set component and component units
       const setComponents = components.slice(idx, idx + 2);
       const setComponentAddresses = _.map(setComponents, token => token.address);
-      const setComponentUnits = _.map(setComponents, () => naturalUnit.mul(idx + 1)); // Multiple of naturalUnit
+      const setComponentUnits: BigNumber[] =
```
felix2feng: Improve formatting / readability by putting braces on their own lines

utils/RebalancingTokenWrapper.ts
```diff
@@ -47,27 +49,47 @@ export class RebalancingTokenWrapper {
...
+      [new BigNumber(10 ** (decOne.toNumber() - minDec)).mul(new BigNumber(idx + 1)),
```
felix2feng: Can we have simpler / intermediate calculations for component units?

utils/RebalancingTokenWrapper.ts
bweick: Exponent of the BigNumber

test/core/rebalancingSetToken.spec.ts
```diff
@@ -859,7 +838,7 @@ contract('RebalancingSetToken', accounts => {
         });
       });
 
-      describe('but the proposed rebalancingSet is not approved by Core', async () => {
+      describe('but the proposed nextSet is not approved by Core', async () => {
```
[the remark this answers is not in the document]
bweick: I would look at it in the context of the whole statement being printed out "when propose is called from the Default state"..."but"

PR #217, opened 2018-09-27
Only transfer maker token specified by signed issuance order

asoong:
We always transfer the maker token specified in the signed issuance order instead of one passed in through the orders data exchange header which can be arbitrary.

### DIFF
--- a/contracts/core/extensions/CoreIssuanceOrder.sol
+++ b/contracts/core/extensions/CoreIssuanceOrder.sol
@@ -115,7 +115,7 @@ contract CoreIssuanceOrder is
                 order.makerAddress,
                 _v,
                 sigBytes[0], // r
-                sigBytes[1] // s
+                sigBytes[1]  // s
             )
         );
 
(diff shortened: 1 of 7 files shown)


contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -268,17 +269,18 @@ contract CoreIssuanceOrder is
...
                 scannedBytes.add(exchangeDataLength)
             );
 
-            // Transfer header.makerTokenAmount to Exchange Wrapper
+            // Transfer maker token to Exchange Wrapper to execute exchange orders
+            // Using maker token from signed issuance order to prevent malicious encoding of another maker token
             ITransferProxy(state.transferProxy).transfer(
-                header.makerTokenAddress,
+                _makerTokenAddress,
                 header.makerTokenAmount,
```
asoong: The exchange header still specifies how much of the maker token to send to each wrapper to execute the orders/trades, but it must be of the maker token from the signed issuance order. If it is the wrong erc20 token, the wrapper will not have enough to execute the trades.

utils/orders.ts
```diff
@@ -74,36 +73,14 @@ export async function generateFillOrderParameters(
   };
 }
 
-export function generateOrdersDataForOrderCount(
```
felix2feng: Assuming this is unused?

asoong:
Yes

PR #218, opened 2018-09-27
Exchange wrapper optimizations, minimize unnecessary transfers/approvals

asoong:
Passing in additional parameters into the wrappers. This will require a redeploy of contracts for it to work on TestNet since the wrappers and parameters that core pass are no longer sufficient.

- Kyber wrapper does one approval for the source token
- Kyber wrapper transfers source token 'change' to the issuance order maker once. This was actually a critical bug
- ZeroEx wrapper does one approval for the issuance order taker token

### DIFF
--- a/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
+++ b/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
@@ -33,8 +33,8 @@ import { LibBytes } from "../../external/0x/LibBytes.sol";
 contract KyberNetworkWrapper is
     Authorizable
 {
-    using SafeMath for uint256;
     using LibBytes for bytes;
+    using SafeMath for uint256;
 
     /* ============ State Variables ============ */
 
@@ -79,16 +79,20 @@ contract KyberNetworkWrapper is
      * We currently pass change back to the issuance order maker, exploring how it can safely be passed to the taker.
      *
      *
-     * @param  _maker         Address of issuance order signer to conform to IExchangeWrapper
-     * @param  _              Unused address of fillOrder caller to conform to IExchangeWrapper
-     * @param  _tradeCount    Amount of trades in exchange request
-     * @param  _tradesData    Byte string containing (multiple) Kyber trades
-     * @return address[]      Array of token addresses traded for
-     * @return uint256[]      Array of token amounts traded for
+     * @param  _maker               Address of issuance order signer to conform to IExchangeWrapper
+     * -- Unused address of fillOrder caller to conform to IExchangeWrapper --
+     * @param  _makerToken          Address of maker token used in exchange orders
+     * @param  _makerAssetAmount    Amount of issuance order maker token to use on this exchange
+     * @param  _tradeCount          Amount of trades in exchange request
+     * @param  _tradesData          Byte string containing (multiple) Kyber trades
+     * @return address[]            Array of token addresses traded for
+     * @return uint256[]            Array of token amounts traded for
      */
     function exchange(
         address _maker,
-        address _,
+        address,
+        address _makerToken,
+        uint256 _makerAssetAmount,
         uint256 _tradeCount,
         bytes _tradesData
     )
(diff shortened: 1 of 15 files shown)


contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -79,105 +79,107 @@ contract KyberNetworkWrapper is
...
+            address(this),
+            kyberNetworkProxy,
+            _makerAssetAmount
+        );
+
+        address[] memory componentTokensReceived = new address[](_tradeCount);
+        uint256[] memory componentTokensAmounts = new uint256[](_tradeCount);
 
-        uint256 scannedBytes = 0;
+        // Prase and execute the trade at the current offset via the KyberNetworkProxy, each kyber trade is 160 bytes
```
bweick: Parse

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -79,105 +79,107 @@ contract KyberNetworkWrapper is
...
+        }
 
-            // Execute the trade via the KyberNetworkProxy
-            takerTokens[i] = trade.destinationToken;
-            takerAmounts[i] = tradeOnKyberReserve(
-                trade,
-                _maker
+        // Transfer any unused or remainder maker token back to the issuance order user
+        uint remainderSourceToken = ERC20.balanceOf(_makerToken, this);
+        if (remainderSourceToken > 0) {
```
felix2feng: ⚠️ If the `componentTokensReceived` during the trade was the `makerToken`, then an unintended amount would be returned to the _maker and the issuance order would fail.

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -79,105 +79,107 @@ contract KyberNetworkWrapper is
...
-        uint remainderSourceToken = ERC20.balanceOf(_trade.sourceToken, this);
-        if (remainderSourceToken > 0) {
-            ERC20.transfer(
-                _trade.sourceToken,
-                _maker,
-                remainderSourceToken
-            );
-        }
-
         // Ensure the maker token is allowed to be transferred by Set TransferProxy
```
felix2feng: `Destination token` vs maker token?

contracts/core/exchange-wrappers/ZeroExExchangeWrapper.sol
```diff
@@ -71,117 +71,118 @@ contract ZeroExExchangeWrapper is
...
         uint256 _orderCount,
         bytes _ordersData
     )
         external
         onlyAuthorized
         returns (address[], uint256[])
     {
-        address[] memory takerTokens = new address[](_orderCount);
-        uint256[] memory takerAmounts = new uint256[](_orderCount);
+        // Ensure the taker token is allowed to be transferred by ZeroEx Proxy
```
felix2feng: I know what you mean, but this is a slightly confusing comment given that the `_makerToken` is passed in.

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
felix2feng: adding comment to request changes

asoong: why is that a problem? why wouldn't we want it to revert? this seems like an optimization to make at the assertion level

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
asoong: +1

contracts/core/exchange-wrappers/ZeroExExchangeWrapper.sol
asoong: +1

felix2feng:
Compile issues 😢

PR #220, opened 2018-09-28
Integrate AuctionPriceCurve into RebalancingSetToken.

bweick:


test/core/extensions/coreRebalanceAuction.spec.ts
```diff
@@ -79,6 +81,7 @@ contract('CoreRebalanceAuction', accounts => {
     coreMock = await coreWrapper.deployCoreMockAsync(transferProxy, vault);
     factory = await coreWrapper.deploySetTokenFactoryAsync(coreMock.address);
     rebalancingFactory = await coreWrapper.deployRebalancingSetTokenFactoryAsync(coreMock.address);
+    constantAuctionPriceCurve = await coreWrapper.deployConstantAuctionPriceCurveAsync(DEFAULT_AUCTION_PRICE);
```
asoong: move to `rebalancingWrapper.ts`

bweick: Only reason I'd be hesitant to do this is because at this point we run all our contract deployment out of coreWrapper. Not sure there's a great reason to change that now just because it's something related to rebalancing and we have a rebalancingWrapper. (Fine with changing the name of the rebalancingWrapper file though)

asoong: that's not true, we deploy exchange wrappers via the ExchangeWrapper and tokens via erc20Wrapper. we should should make the change as soon as it makes sense, and slowly migrate existing functions that don't make sense over time. (e.g. deployRebalancingSetTokenAsync should probably be in reblanacingWrapper) as well to create distinction between it and createRebalancingTokenAsync. these are all marginal changes that improves someones ability to get into navigate the code source quickly

PR #222, opened 2018-09-29
Alex/kyber wrapper issuance specs

asoong:
This required several updates to the snapshot including increasing the reserve balance, category cap, and imbalances.

- Added a kyber trade as one of the liquidity sources in the core issuance happy path spec.
- Added helper method on the contract for getting conversion rates from kyber network so we don't need to use their library/api in ours

### DIFF
--- a/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
+++ b/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
@@ -152,7 +152,7 @@ contract KyberNetworkWrapper is
         }
 
         // Transfer any unused or remainder maker token back to the issuance order user
-        uint remainderMakerToken = ERC20.balanceOf(_makerToken, this);
+        uint256 remainderMakerToken = ERC20.balanceOf(_makerToken, this);
         if (remainderMakerToken > 0) {
             ERC20.transfer(
                 _makerToken,
--- a/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
+++ b/contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
@@ -73,6 +73,38 @@ contract KyberNetworkWrapper is
     /* ============ Public Functions ============ */
 
     /**
+     * Returns the conversion rate between the issuance order maker token and the set component token
+     * in 18 decimals, regardless of component token's decimals
+     *
+     * @param  _makerToken        Address of maker token used in exchange orders
+     * @param  _componentToken    Address of set component token to trade for
+     * @param  _quantity          Amount of maker token to exchange for component token
+     * @return uint256            Conversion rate in wei
+     * @return uint256            Slippage in wei
+     */
+    function conversionRate(
+        address _makerToken,
+        address _componentToken,
+        uint256 _quantity
+    )
+        external
+        returns (uint256, uint256)
+    {
+        uint256 rate;
+        uint256 slippage;
+        (rate, slippage) = KyberNetworkProxyInterface(kyberNetworkProxy).getExpectedRate(
+            _makerToken,
+            _componentToken,
+            _quantity
+        );
+
+        return (
+            rate,
+            slippage
+        );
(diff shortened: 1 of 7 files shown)


contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -120,12 +152,12 @@ contract KyberNetworkWrapper is
         }
 
         // Transfer any unused or remainder maker token back to the issuance order user
-        uint remainderSourceToken = ERC20.balanceOf(_makerToken, this);
-        if (remainderSourceToken > 0) {
+        uint remainderMakerToken = ERC20.balanceOf(_makerToken, this);
```
felix2feng: Specify `uint256`

contracts/external/KyberNetwork/KyberNetworkProxyInterface.sol
```diff
@@ -3,6 +3,15 @@ pragma solidity 0.4.24;
 
 /// @title Kyber Network interface
 interface KyberNetworkProxyInterface {
+    function getExpectedRate(
+      address src,
+      address dest,
+      uint srcQty
```
felix2feng: Unless its an issue, specify `uint256?`.

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -212,6 +225,26 @@ contract('CoreIssuanceOrder', accounts => {
...
       );
 
+      // Create Kyber trade for the third component, using ether(25) makerToken. Conversion rate pre set on snapshot
+      const sourceTokenQuantity = ether(25);
+      const maxDestinationQuantity = issuanceOrderRequiredComponentAmounts[2];
+      const componentTokenDecimals = (await thirdComponent.decimals.callAsync()).toNumber();
+      const sourceTokenDecimals = (await makerToken.decimals.callAsync()).toNumber();
+      kyberConversionRatePower = new BigNumber(10).pow(18 + sourceTokenDecimals - componentTokenDecimals);
+      const minimumConversionRate = maxDestinationQuantity.div(sourceTokenQuantity)
+                                                          .mul(kyberConversionRatePower)
```
felix2feng: I love the math steps put on their own line

test/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -212,6 +225,26 @@ contract('CoreIssuanceOrder', accounts => {
...
+                                                          .round();
+      kyberTradeMakerTokenChange = sourceTokenQuantity.sub(
+        maxDestinationQuantity.mul(kyberConversionRatePower).div(KYBER_RESERVE_CONFIGURED_RATE).floor());
```
felix2feng: Possible to separate math steps into their own line here?

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
asoong: +1

contracts/external/KyberNetwork/KyberNetworkProxyInterface.sol
asoong: This is in the external folder. Leaving it to match their interface like we did in the other external files

test/core/extensions/coreIssuanceOrder.spec.ts
asoong: +1

test/core/extensions/coreIssuanceOrder.spec.ts
asoong: Not a lot of clarity or meaning for the in between steps in this case

asoong: Oh, I see what you mean, I thought it was confusing cause it's layered inside one sub already so I didn't do that here

PR #227, opened 2018-10-11
Added check to make sure internalDeposit doesn't call zero quantity t…

bweick:
…ransfers.


bweick:
Further question...should this be put on the transferProxy? It would add more gas to the typical txn but would also blanket guarantee no zero value errors. My feeling is we can be judicious enough to not make it a problem...but at the same time this would guarantee no issues.

PR #230, opened 2018-10-12
Web3 1.0 changes

justinkchen:


package.json
```diff
@@ -71,31 +69,34 @@
...
     "husky": "^0.14.3",
     "lint-staged": "^7.2.0",
     "module-alias": "^2.1.0",
-    "set-protocol-utils": "^0.3.37",
+    "set-protocol-utils": "^1.0.0-beta.10",
     "sol-trace-set": "^0.0.1",
     "solium": "^1.1.7",
     "tiny-promisify": "^1.0.0",
-    "truffle-hdwallet-provider": "^0.0.3",
+    "truffle-hdwallet-provider": "^1.0.0-web3one.0",
```
felix2feng: That's cool this dep has a web3 1.0 version

tsconfig.dist.json
```diff
@@ -15,23 +15,25 @@
...
         "node_modules/@0xproject/types/*",
-        "node_modules/@0xproject/typescript-typings/types/*",
+        "node_modules/set-protocol-utils/types/*",
         "node_modules/*"
       ]
     },
     "declaration": true,
-    "declarationDir": "./dist/typings"
+    "declarationDir": "./dist/typings",
+    "esModuleInterop": true
```
felix2feng: What's the benefit / need for `esModuleInterop`?

types/base_contract.ts
```diff
@@ -14,14 +14,14 @@ export class BaseContract {
     public abi: any[];
     // public abi: Web3.AbiDefinition[];
 
-    public web3ContractInstance: Web3.ContractInstance;
+    public web3ContractInstance: any;
```
felix2feng: I get really nervous when we're moving to an `any` type. Is there another type we could use?

felix2feng: If it doesn't exist, we should consider creating our own type

tsconfig.dist.json
asoong: https://www.typescriptlang.org/docs/handbook/compiler-options.html

truffle.js
```diff
@@ -38,7 +38,15 @@ module.exports = {
       gas: 0xfffffffffff,
       gasPrice: 0x01,
     },
-  }
+  },
+  compilers: {
```
asoong: do we still need the top level `solc:` option?

package.json
justinkchen: it is required as part of the truffle 5 migration efforts too i think

tsconfig.dist.json
justinkchen: Required in order to allow for default imports (so we don't have to do `require('...')`) for modules that don't explicitly have that

types/base_contract.ts
justinkchen: There is no other type right now defined at least.  `web3.eth.Contract` is a class that has a constructor to generate classes, but I couldn't get it working as a type yet

truffle.js
justinkchen: i think we may be able to get rid of that

types/base_contract.ts
justinkchen: i guess same thing should be done for the `// public abi: Web3.AbiDefinition[];` comment we have above. not sure why it was commented to be `any` there as well

truffle.js
justinkchen: interesting, i know it picks up the compiler version, but maybe the solc is not somehow...or maybe its in your local truffle version not using it properly? Didn't you say you weren't on truffle 5 yet?

justinkchen: we can also keep for backwards compatibility on older truffle versions

tsconfig.dist.json
justinkchen: in particular i think web3 and some other package had this problem that this was able to solve

PR #234, opened 2018-10-17
Passing fees through core and onto the rebalancing set token

asoong:


test/core/rebalancingSetToken.spec.ts
```diff
@@ -478,6 +513,8 @@ contract('RebalancingSetToken', accounts => {
       const initialUnitShares = DEFAULT_UNIT_SHARES;
       const proposalPeriod = ONE_DAY_IN_SECONDS;
       const rebalanceInterval = ONE_DAY_IN_SECONDS;
+      const entranceFee = ZERO;
```
bweick: Might be a bit of a nit but feel like these should be initialized with a non-zero value, because if nothing is assigned to the state var on the smart contract I believe it'll return 0. So I think to avoid an unlikely, though possible regression (the code looks fine) we should make this a positive number.

test/core/rebalancingSetToken.spec.ts
```diff
@@ -128,6 +130,8 @@ contract('RebalancingSetToken', accounts => {
       subjectInitialUnitShares = DEFAULT_UNIT_SHARES;
       subjectProposalPeriod = ONE_DAY_IN_SECONDS;
       subjectRebalanceInterval = ONE_DAY_IN_SECONDS;
+      subjectEntranceFee = ZERO;
```
bweick: Might be a bit of a nit but feel like these should be initialized with a non-zero value, because if nothing is assigned to the state var on the smart contract I believe it'll return 0. So I think to avoid an unlikely, though possible regression (the code looks fine) we should make this a positive number.

test/core/rebalancingSetTokenFactory.spec.ts
```diff
@@ -85,100 +86,87 @@ contract('RebalancingSetTokenFactory', accounts => {
...
-        SetUtils.paddedBufferForBigNumber(proposalPeriod),
-        SetUtils.paddedBufferForBigNumber(rebalanceInterval),
-      ]);
+      subjectName = 'My Rebalancing Set';
+      subjectSymbol = 'REBAL';
+
+      callDataManagerAddress = rebalancingTokenManagerAccount;
+      callDataProposalPeriod = new BigNumber(86400);
+      callDataRebalanceInterval = new BigNumber(86400);
+      callDataEntranceFee = ZERO;
```
bweick: Same as above here

test/core/rebalancingSetToken.spec.ts
asoong: +1

PR #236, opened 2018-10-18
Brian/add fees on rebalancing

bweick:


contracts/core/Core.sol
```diff
@@ -61,5 +61,8 @@ contract Core is
 
         // Commit passed address to vault state variable
         state.vault = _vault;
+
+        // Set feesEnabled to false
+        state.feesEnabled = false;
```
asoong: Low cost to us at deploy time but I think it defaults to false

contracts/core/Core.sol
```diff
@@ -61,5 +61,8 @@ contract Core is
 
         // Commit passed address to vault state variable
         state.vault = _vault;
+
+        // Set feesEnabled to false
```
asoong: "Initially disable protocol fee to false"

contracts/core/RebalancingSetToken.sol
```diff
@@ -269,12 +271,20 @@ contract RebalancingSetToken is
         // Make sure all currentSets have been rebalanced
         require(remainingCurrentSets < minimumBid, "REBALANCE_NOT_FINISHED");
 
-        // Creating pointer to Core to Issue next set and Deposit into vault
+        // Creating pointer to Core to Issue next set and Deposit into vault and to nextSet token
+        // to transfer fees
         ICore core = ICore(ISetFactory(factory).core());
+        ISetToken nextSetInstance = ISetToken(nextSet);
+        address protocolAddress = ICore(ISetFactory(factory).core()).protocolAddress();
```
asoong: I dont see a protocol address method on core, is it somewhere else?

contracts/core/extensions/CoreInternal.sol
```diff
@@ -97,4 +97,30 @@ contract CoreInternal is
...
+        onlyOwner
+    {
+        // Find and remove factory from factories array
+        state.protocolAddress = _protocolAddress;
+    }
+
+    /**
+     * Turn protocol fees on for collecting rebalancing fees
+     */
+    function enableFees()
```
asoong: Can we make this a bi directional setter? In case it cucks something and we want to turn it back off?

contracts/core/extensions/CoreInternal.sol
```diff
@@ -97,4 +97,30 @@ contract CoreInternal is
...
+        external
+        onlyOwner
+    {
+        // Set feesEnabled to true
+        state.feesEnabled = true;
```
asoong: Instead of true or false, I think we want this to be a rate that begins as `0`

contracts/core/RebalancingSetToken.sol
asoong: Found it

contracts/core/RebalancingSetToken.sol
```diff
@@ -291,9 +301,21 @@ contract RebalancingSetToken is
...
+            managerFee
+        );
+
+        if (protocolFee > 0) {
+            nextSetInstance.transfer(
+                protocolAddress,
+                protocolFee
+            );
+        }
+
```
asoong: Don't we need to update the unit shares after we transfer all the fees out? There is less withdrawable for everyone now

asoong: Oh I see you did that in calculateNextSetIssueQuantity

contracts/core/extensions/CoreInternal.sol
justinkchen: what was the reasoning behind not doing it this way?

PR #246, opened 2018-10-25
Add price lib tracking in Core

felix2feng:
Completes this [card](https://trello.com/c/TyeL17Lc/630-coreinternal-new-functions).

Add new functions: addPriceLibrary and removePriceLibrary so that only approved price libraries can be used.

### DIFF
--- a/contracts/core/extensions/CoreInternal.sol
+++ b/contracts/core/extensions/CoreInternal.sol
@@ -105,6 +105,47 @@ contract CoreInternal is
     }
 
     /**
+     * Add a price library to the mapping of tracked price libraries. Can only be set by
+     * owner of Core.
+     *
+     * @param  _priceLibrary   The address of the Price Library to enable
+     */
+    function enablePriceLibrary(
+        address _priceLibrary
+    )
+        external
+        onlyOwner
+    {
+        // Mark as true in validPriceLibraries mapping
+        state.validPriceLibraries[_priceLibrary] = true;
+
+        // Add to priceLibraries array
+        state.priceLibraries.push(_priceLibrary);
+    }
+
+    /**
+     * Disable a priceLibrary in the mapping of tracked Price Libraries. Can only be disabled
+     * by owner of Core.
+     *
+     * @param  _priceLibrary   The address of the Price Library to disable
+     */
+    function disablePriceLibrary(
+        address _priceLibrary
+    )
+        external
+        onlyOwner
+    {
+        // Verify Library is linked to Core
+        require(state.validPriceLibraries[_priceLibrary], "UNKNOWN_LIBRARY");
+
+        // Mark as false in validPriceLibraries mapping
+        state.validPriceLibraries[_priceLibrary] = false;
+
+        // Find and remove price library from priceLibraries array
+        state.priceLibraries = state.priceLibraries.remove(_priceLibrary);
+    }
+
+    /**
      * Change address that rebalancing protocol fees accrue to
      *
      * @param  _protocolAddress   The protcol fee address
--- a/contracts/core/extensions/CoreInternal.sol
+++ b/contracts/core/extensions/CoreInternal.sol
@@ -105,23 +105,25 @@ contract CoreInternal is
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/core/extensions/CoreInternal.sol
```diff
@@ -104,6 +104,47 @@ contract CoreInternal is
...
+        state.priceLibraries.push(_priceLibrary);
+    }
+
+    /**
+     * Disable a priceLibrary in the mapping of tracked Price Libraries. Can only be disabled
+     * by owner of Core.
+     *
+     * @param  _priceLibrary   The address of the Price Library to disable
+     */
+    function disablePriceLibrary(
```
asoong: Why do we also need a disable method? Can we instead have one setter method where we set whether a price library is enabled?

felix2feng: We could. I was just following the pattern we already have in place.

asoong: Yeah it doesn't make sense to have two methods, lets add a card to get the other ones changed as well (also making `disableSet` bi directional). I'm also not sure we need the array and the map.

felix2feng: https://trello.com/c/3pNq8JyQ/640-refactor-coreinternal-add-remove-functions-into-a-single-setter-function

contracts/core/interfaces/ICore.sol
```diff
@@ -125,6 +125,17 @@ interface ICore {
     )
         external;
 
+    /**
+     * Adds or removes a price library to the mapping of tracked price libraries. Can only be set by
+     * owner of Core.
+     *
+     * @param  _priceLibrary   The address of the PriceLibrary to enable/disable
+     */
+    function setPriceLibrary(
```
asoong: Why are we exposing this/other setters again? Not sure we ever need to expose the owner-only setters. I think this adds to the runtime gas cost by having them all in a single core interface. @bweick, we should do an exercise on that and confirm. If true, we should definitely have logic specific interfaces that expose 1-2 methods at a time.

contracts/core/extensions/CoreInternal.sol
```diff
@@ -104,6 +104,30 @@ contract CoreInternal is
         state.setTokens = state.setTokens.remove(_set);
     }
 
+    /**
+     * Adds or removes a price library to the mapping of tracked price libraries. Can only be set by
+     * owner of Core.
+     *
+     * @param  _priceLibrary   The address of the Price Library to enable or disable
+     */
+    function setPriceLibrary(
```
asoong: `setPriceLibraryValid(address _priceLibrary, boolean _valid)` or `setPriceLibraryEnabled(address _priceLibrary, boolean _enabled)`?

contracts/core/extensions/CoreInternal.sol
```diff
@@ -104,6 +104,30 @@ contract CoreInternal is
...
+        address _priceLibrary
+    )
+        external
+        onlyOwner
+    {
+        // Mark as true or false in validPriceLibraries mapping
+        state.validPriceLibraries[_priceLibrary] = !state.validPriceLibraries[_priceLibrary];
+
+        if (state.validPriceLibraries[_priceLibrary]) {
```
asoong: Then you can save on a lot of lookups here, not that this is a function we're going to be calling often.

contracts/core/extensions/CoreInternal.sol
```diff
@@ -104,6 +104,30 @@ contract CoreInternal is
...
+     * @param  _priceLibrary   The address of the Price Library to enable or disable
+     */
+    function setPriceLibrary(
+        address _priceLibrary
+    )
+        external
+        onlyOwner
+    {
+        // Mark as true or false in validPriceLibraries mapping
+        state.validPriceLibraries[_priceLibrary] = !state.validPriceLibraries[_priceLibrary];
```
asoong: Same, a ton of lookups there. I'll be using one of the interfaces above for the other methods.

PR #249, opened 2018-10-25
Add validateOrder assertions and unit tests

felix2feng:


contracts/core/RebalancingSetToken.sol
```diff
@@ -578,6 +578,21 @@ contract RebalancingSetToken is
         return units;
     }
 
+    /*
+     * Checks to make sure token is component of Set
+     *
+     * @param  _tokenAddress     Address of token being checked
+     * @return  bool             True if token is component of Set
+     */
+    function tokenIsComponent(
```
asoong: Where do you intend on using this check again? Can the caller retrieve the current set from the token since its only a single selector?

If you still think its relevant, rename to something more appropriate; using the component/unit schema to create rebalancing sets doesn't mean we have to conform to the nomenclature for other methods. Maybe: `isCurrentSet`?

felix2feng: Its used below in the issuanceOrder require checks. 

In this case, we need a common method for checking whether a component is part of a Set.

asoong: This is fine then after we discussed yesterday, can you update the comment/param descriptions to be more effective.

`Checks to make sure address is the current set of the RebalancingSetToken. Conforms to the ISetToken Interface.`

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -328,8 +328,13 @@ contract CoreIssuanceOrder is
...
         view
     {
-        //Declare set interface variable
+        // Declare set interface variable
         ISetToken set = ISetToken(_order.setAddress);
+        address[] memory requiredComponents = _order.requiredComponents;
+        uint256[] memory requiredComponentAmounts = _order.requiredComponentAmounts;
+
+        // Make sure maker token addressesis non-null
+        require(_order.makerToken != address(0), "ORDER_MAKER_TOKEN_INVALID");
```
asoong: Why is this necessary?

felix2feng: You can create an issuance order with a null makerToken with a positive makerTokenAmount - which doesn't make sense.

But a non-null makerToken with a 0 amount makes sense (if we're trying to describe a charity-related use case).

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -328,8 +328,13 @@ contract CoreIssuanceOrder is
         private
         view
     {
-        //Declare set interface variable
+        // Declare set interface variable
         ISetToken set = ISetToken(_order.setAddress);
+        address[] memory requiredComponents = _order.requiredComponents;
+        uint256[] memory requiredComponentAmounts = _order.requiredComponentAmounts;
+
+        // Make sure maker token addressesis non-null
```
asoong: `address is`

contracts/core/extensions/CoreIssuanceOrder.sol
asoong: Is it a security issue? This seems more like an assertion on the setProtocol.js level. The IssuanceOrder cannot be completed if the maker token is null addressed

felix2feng: True, I don't think its a security issue - since null address doesn't have transfer. But it saves users gas (by not attempting transfers) who are potentially calling with a null maker token.

asoong:
I removed the null address requires but i'm not sure any of these are required. Hold off on merging this PR for now.

contracts/core/extensions/CoreIssuanceOrder.sol
```diff
@@ -307,6 +309,23 @@ contract CoreIssuanceOrder is
 
         // Make sure fill or cancel quantity is multiple of natural unit
         require(_executeQuantity % set.naturalUnit() == 0, "FILL_NOT_NATURAL_UNIT_MULTIPLE");
+
+        // Make sure required components array is non-empty
+        require(_order.requiredComponents.length > 0, "ORDER_REQ_COMPONENTS_EMPTY");
```
hydrazines: We updated our error messages to a new format that these should be updated to as well.

hydrazines:
@asoong Updated thoughts on if these checks should happen? I don't think I have enough context to make a call.

test/contracts/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -778,6 +780,63 @@ contract('CoreIssuanceOrder', accounts => {
...
+        await expectRevertError(subject());
+      });
+    });
+
+    describe('when a required component amount is 0', async () => {
+      before(async () => {
+        issuanceOrderRequiredComponentAmounts = [ZERO, ZERO, ZERO];
+      });
+
+     after(async () => {
```
hydrazines: Need one more space before `after`

test/contracts/core/extensions/coreIssuanceOrder.spec.ts
```diff
@@ -778,6 +780,63 @@ contract('CoreIssuanceOrder', accounts => {
...
+        await expectRevertError(subject());
+      });
+    });
+
+    describe('when the required components and amount lengths differ', async () => {
+      before(async () => {
+        issuanceOrderRequiredComponents = [notIssuanceOrderMaker];
+      });
+
+     after(async () => {
```
hydrazines: Same here

PR #250, opened 2018-10-25
Add rebalancingSetToken create assertions

felix2feng:
Completes this [card](https://trello.com/c/sWeioqsq/626-rebalancingsettokenfactorysol-add-checks-and-modifiers).

### DIFF
--- a/contracts/core/RebalancingSetTokenFactory.sol
+++ b/contracts/core/RebalancingSetTokenFactory.sol
@@ -99,6 +99,9 @@ contract RebalancingSetTokenFactory {
         // Expecting caller to be Core
         require(msg.sender == core, "ONLY_CORE_CAN_CREATE_REBAL_SET");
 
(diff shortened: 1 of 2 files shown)


contracts/core/RebalancingSetTokenFactory.sol
```diff
@@ -99,6 +99,9 @@ contract RebalancingSetTokenFactory {
         // Expecting caller to be Core
         require(msg.sender == core, "ONLY_CORE_CAN_CREATE_REBAL_SET");
 
+        // Ensure components and units length are 1
+        require(_components.length == 1 && _units.length == 1, "ARRAY_LENGTHS_MUST_BE_ONE");
```
asoong: Can you make this message uniquely identifiable but < 32 characters? I recommend we change these around to something like: `RebalancingFactory#Const.ArrayLength` @bweick

PR #253, opened 2018-10-30
Add rebalancingSet/propose checks

felix2feng:
Completes this [card](https://trello.com/c/tC8vfU0w/628-rebalancingsettoken-propose-additional-assert-statement-add-library-check)

### DIFF
--- a/contracts/core/RebalancingSetToken.sol
+++ b/contracts/core/RebalancingSetToken.sol
@@ -207,10 +207,10 @@ contract RebalancingSetToken is
         // Check that the auction library is a valid priceLibrary tracked by Core
         require(core.validPriceLibraries(_auctionLibrary), "PRICE_LIB_MUST_BE_VALID");
 
-        // Assert price divisor > 0
+        // Assert price divisor is non-zero, ensuring a positive slope
         require(_auctionPriceDivisor > 0, "PRICE_DIV_MUST_BE_NON_ZERO");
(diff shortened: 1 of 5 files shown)


contracts/core/RebalancingSetToken.sol
```diff
@@ -199,9 +200,18 @@ contract RebalancingSetToken is
...
-
+        
         // Check that new proposed Set is valid Set created by Core
-        require(ICore(ISetFactory(factory).core()).validSets(_nextSet), "PROPOSED_SET_INVALID");
+        require(core.validSets(_nextSet), "PROPOSED_SET_INVALID");
+
+        // Check that the auction library is a valid priceLibrary tracked by Core
+        require(core.validPriceLibraries(_auctionLibrary), "PRICE_LIB_MUST_BE_VALID");
+
+        // Assert price divisor > 0
```
asoong: These are really ineffective comments, provide some color here on why we do this

contracts/core/RebalancingSetToken.sol
```diff
@@ -199,9 +200,18 @@ contract RebalancingSetToken is
...
+        require(_auctionPriceDivisor > 0, "PRICE_DIV_MUST_BE_NON_ZERO");
+
+        // Assert curve coefficient > 0
```
asoong: Same

PR #254, opened 2018-10-31
Combine enabling and disabling factories

asoong:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -50,44 +50,28 @@ contract CoreInternal is
...
      *
-     * @param  _factory   The address of the SetTokenFactory to disable
+     * @param  _factory   Address of the contract conforming to ISetFactory
+     * @param  _enabled   Enable or disable the factory
      */
-    function disableFactory(
-        address _factory
+    function registerFactory(
+        address _factory,
+        bool _enabled
```
felix2feng: This is problematic potentially, because you can call register multiple times and thus add multiple instances of the same address/factory in the array.

I think the bool `enabled` is not needed as it is implicit. If it exists, then you disable it. If not, then you add it.

asoong: The first one is definitely a legitimate issue pre-this PR as well. +1 to that. I'm actually not sure why we keep this array since we only need the mapping for verification. It's inconsistent with our store for exchange wrappers and we've gone on fine without that list.

I'm not a fan of inversing the existing state. The clarity will permeate through to tests where it would be unclear what the current state is after an arbitrary number of `registerFactory` calls, and when we are looking back at transaction hashes.

felix2feng: The array makes sense to get a list of all the available price libraries on-chain. otherwise, the only way to check is to look @ our documentation. That said, I'm not against removing the array.

felix2feng: Also, there is more security for the end users if they can find a list of all enabled factories/libraries. (They would just need to trust us otherwise).
This is due to the impossibility of exploring the entire space to find which addresses are true/false.

asoong: More security, how? I can't imagine a flow where a call to `factories`. I think when I lobbied for individual `set` arrays on the factories, we agreed that that's what logs are for. That same thinking should apply here

felix2feng: I think it would be a security risk for users if they can't validate/verify what factories/libraries are active or enabled. 

Event logs could potentially be a way to check that, but we don't have that implemented. Array list works well too.

contracts/core/extensions/CoreInternal.sol
```diff
@@ -36,12 +36,19 @@ contract CoreInternal is
...
     /* ============ Events ============ */
 
-    // Logs registration of new exchange
+    // Logs registration of new exchange conforming to IExchangeWrapper
     event ExchangeRegistered(
         uint8 _exchangeId,
         address _exchange
     );
 
+    // Logs registration of a new factory conforming to ISetFactory
```
hydrazines: Nit: More accurate to say something like "Logs a change in the registration of a factory..."

contracts/core/extensions/CoreInternal.sol
```diff
@@ -50,44 +57,25 @@ contract CoreInternal is
...
-        // Verify Factory is linked to Core
-        require(state.validFactories[_factory], "UNKNOWN_FACTORY");
-
-        // Mark as false in validFactories mapping
-        state.validFactories[_factory] = false;
+        state.validFactories[_factory] = _enabled;
 
-        // Find and remove factory from factories array
-        state.factories = state.factories.remove(_factory);
+        emit FactoryRegistrationChanged(
```
hydrazines: This means an event will be fired on each method call, even if the state did not change. Do we want to fire an event only on explicit state change?

test/contracts/core/extensions/coreInternal.spec.ts
```diff
@@ -98,11 +105,19 @@ contract('CoreInternal', accounts => {
       expect(isFactoryValid).to.be.true;
     });
 
-    it('adds setTokenFactory address to factories array correctly', async () => {
-      await subject();
+    it('emits a FactoryRegistrationChange event', async () => {
```
hydrazines: Nit: should be `FactoryRegistrationChanged`, same with other occurrence

test/contracts/core/extensions/coreInternal.spec.ts
```diff
@@ -20,7 +20,11 @@ import {
 import { expectRevertError } from '@utils/tokenAssertions';
 import { Blockchain } from '@utils/blockchain';
 import { STANDARD_NATURAL_UNIT } from '@utils/constants';
-import { getExpectedFeeStatusChangeLog, ExchangeRegistered } from '@utils/contract_logs/core';
+import {
+  getExpectedFeeStatusChangeLog,
+  ExchangeRegistered,
+  FactoryRegistrationChanged
```
justinkchen: nit: trailing comma

contracts/core/extensions/CoreInternal.sol
justinkchen: how does this add multiple factories to an array? isnt this a mapping of factoryAddress => boolean?

felix2feng: @justinkchen It doesn't anymore with the latest commit. We previously had a factories array

contracts/core/extensions/CoreInternal.sol
asoong: Ambivalent here, the logs and the map will reflect the last state. I argue for these in particular that it's merely to track that the event was called so we can always build that array that we removed

PR #255, opened 2018-10-31
Upgrade pricing library registration

asoong:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -142,7 +142,7 @@ contract CoreInternal is
      * @param  _priceLibrary   Address of contract Price Library to enable or disable
      * @param  _enabled        Whether the pricing library is enabled for use in proposal cycle
      */
-    function setPriceLibraryEnabled(
+    function registerPriceLibrary(
         address _priceLibrary,
         bool _enabled
```
felix2feng: Same comment. Why are we adding the `bool` enabled? Why do we need it?

asoong: https://github.com/SetProtocol/set-protocol-contracts/pull/254

felix2feng: Cool - as long as we have an event emitted, this works

contracts/core/extensions/CoreInternal.sol
```diff
@@ -54,6 +54,12 @@ contract CoreInternal is
         bool _newStatus
     );
 
+    // Logs pricing library registration change. Library must conform to IAuctionPriceCurve
+    event PricingLibraryRegistrationChanged(
```
hydrazines: Nit: Change to `PriceLibraryRegistrationChanged` for consistency with function

contracts/core/extensions/CoreInternal.sol
```diff
@@ -54,6 +54,12 @@ contract CoreInternal is
...
+        address _pricingLibrary,
```
hydrazines: Same here

PR #258, opened 2018-11-01
Refactor #disableSet to #registerSet

hydrazines:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -114,25 +120,39 @@ contract CoreInternal is
...
      *
-     * @param  _set   The address of the SetToken to disable
+     * @param  _set       The address of the Set
+     * @param  _enabled   Enable or disable the Set
      */
-    function disableSet(
-        address _set
+    function registerSet(
+        address _set,
+        bool _enabled
```
felix2feng: @asoong The ability for us to register Sets gives us very straightforward access to create a Set that can drain our vault. For example, we can deploy a token that conforms to the Set interface (made of various components) that has an existing supply attached to it (even though we never called issue through the system). Then, we can register it using this function and then call redeem without having contributed components.

Our powers as operators of the platform should be generally that of adding new features (e.g. factory, exchange wrapper, etc.) or putting the protocol in an emergency state in the event things go wrong. Thus, allowing administrators to add an arbitrary Set w/o going through a factory seems inconsistent with that.

That said, I think there is a place for a registerSet function that can enable/disable Sets.
When a Set has been disabled, it would make sense for it to be re-enabled - so that the fund's stored by the Set aren't retrievable by the vault forever.

One solution is to track Sets that has been disabled in a mapping, maybe called `disabledSets`. And enabling is only allowed/restricted to disabled Sets.

test/contracts/core/extensions/coreInternal.spec.ts
```diff
@@ -20,7 +20,11 @@ import {
 import { expectRevertError } from '@utils/tokenAssertions';
 import { Blockchain } from '@utils/blockchain';
 import { STANDARD_NATURAL_UNIT } from '@utils/constants';
-import { getExpectedFeeStatusChangeLog, ExchangeRegistered } from '@utils/contract_logs/core';
+import {
+  getExpectedFeeStatusChangeLog,
+  ExchangeRegistered,
+  SetRegistrationChanged
```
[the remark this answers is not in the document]
hydrazines: :+1:

contracts/core/extensions/CoreInternal.sol
hydrazines: Noted; as per Slack we're going to punt discussion on this for later as the issue still exists in the case of being able to register arbitrary factories.

PR #260, opened 2018-11-02
Move constructOrder to OrderLibrary, add require to validateSignature…

bweick:
… fxn instead of wrapped around it. Some formatting changes.

### DIFF
--- a/contracts/core/lib/OrderLibrary.sol
+++ b/contracts/core/lib/OrderLibrary.sol
@@ -141,7 +141,52 @@ library OrderLibrary {
             _s
         );
 
-        return recAddress == _signerAddress;
+        require(recAddress == _signerAddress, "INVALID_SIGNATURE");
+    }
+
+    /**
+     * Construct issuance order struct
+     *
+     * @param  _addresses                   [setAddress, makerAddress, makerToken, relayerAddress, relayerToken]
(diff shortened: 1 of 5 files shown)


contracts/core/lib/OrderLibrary.sol
```diff
@@ -141,7 +141,52 @@ library OrderLibrary {
...
+        require(recAddress == _signerAddress, "INVALID_SIGNATURE");
+    }
+
+    /**
+     * Construct issuance order struct
+     *
+     * @param  _addresses                   [setAddress, makerAddress, makerToken, relayerAddress, relayerToken]
+     * @param  _values                      [quantity, makerTokenAmount, expiration, makerRelayerFee, takerRelayerFee, salt]
+     * @param  _requiredComponents          Components to be acquired by exchange order
+     * @param  _requiredComponentAmounts    Amounts of each component to be acquired by exchange order
```
felix2feng: Add return javadoc

utils/tokenAssertions.ts
```diff
@@ -22,6 +22,15 @@ export async function expectRevertError(asyncTxn: any) {
   }
 }
 
+// For solidity function calls that do not violate require()
+export async function expectNoRevertError(asyncTxn: any) {
```
felix2feng: With test cases that pass, I expect there to be an `expect` or test case.

In this case, there is no test case in the try block. Maybe something like expect(true) would suffice if the `await asyncTxn` passes

PR #262, opened 2018-11-05
Brian/add javadoc to interfaces

bweick:


contracts/core/interfaces/IRebalancingSetToken.sol
```diff
@@ -23,13 +23,27 @@ pragma solidity 0.4.24;
  * The IRebalancingSetToken interface provides a light-weight, structured way to interact with the
  * RebalancingSetToken contract from another contract.
  */
+
 interface IRebalancingSetToken {
+    /*
+     * Place bid during rebalance auction. Can only be called by Core.
+     *
+     * @param _quantity                 The amount of currentSet to be rebalanced
+     * @return combinedTokenArray       Array of token addresses invovled in rebalancing
```
asoong: The return type is usually denoted here, more effective than the variable name. So this would be

@return address[]
@return uint256[]
@return uint256[]

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -36,6 +36,7 @@ contract LinearAuctionPriceCurve {
      * @param  _auctionStartTime          Time of auction start
      * @param  _auctionStartPrice         The price to start the auction at
      * @param  _curveCoefficient          The slope (or convexity) of the price curve
+     * @return uint256                    Numerator of calculated price
```
asoong: Like you did here

PR #263, opened 2018-11-06
Update protocol fees enabled to be toggleable amount

asoong:
@bweick how do we guarantee the fee is always a valid divisor?


bweick:
Not sure what exactly you mean...but since fees are in basis points it should always be divided by 10,000

PR #265, opened 2018-11-06
withdrawInternal does not transfer on 0 quantity.

bweick:


contracts/core/extensions/CoreAccounting.sol
```diff
@@ -208,22 +208,25 @@ contract CoreAccounting is
     )
         internal
     {
-        // Declare interface variavle for vault
-        IVault vault = IVault(state.vault);
+        // Don't allow withdraw if no amount
+        if (_quantity != 0) {
```
hydrazines: Should this be `_quantity > 0` to mirror `depositInternal` and disallow negative quantities?

bweick: Definitely should. Dunno why I did that.

PR #267, opened 2018-11-06
Mirror deposit and withdraw internal function patterns

hydrazines:


contracts/core/extensions/CoreAccounting.sol
```diff
@@ -74,7 +74,10 @@ contract CoreAccounting is
         external
         nonReentrant
     {
+        // Call internal withdraw function
         withdrawInternal(
+            msg.sender,
+            msg.sender,
```
hydrazines: Pass `msg.sender` to `withdrawInternal` just like with `depositInternal`

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -117,31 +120,22 @@ contract CoreAccounting is
...
-                _quantities[i]
-            );
-        }
+        // Call internal batch withdraw function
+        batchWithdrawInternal(
+            msg.sender,
+            msg.sender,
+            _tokens,
+            _quantities
+        );
```
hydrazines: Moved the existing batch withdraw to an internal function and call that function here

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -211,30 +211,30 @@ contract CoreAccounting is
         // Declare interface variavle for vault
         IVault vault = IVault(state.vault);
 
-        // Call Vault contract to deattribute tokens to user
-        vault.decrementTokenOwner(
+        // Call Vault contract to withdraw tokens from Vault to user
+        vault.withdrawTo(
```
asoong: What was the thinking behind this? We have it in the existing order for a reason, we update the state before we perform the actual transfer

asoong: Rule of thumb

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -211,30 +211,30 @@ contract CoreAccounting is
...
             _token,
-            msg.sender,
+            _to,
             _quantity
         );
 
-        // Call Vault to withdraw tokens from Vault to user
-        vault.withdrawTo(
+        // Call Vault contract to deattribute tokens from user
+        vault.decrementTokenOwner(
```
hydrazines: Switched the order of `withdrawTo` and `decrementTokenOwner` because, like in deposit, we want to ensure the deposit/withdrawal goes through successfully first, and then increment/decrement the user.

hydrazines: Nevermind, gonna switch this back, I see the `require` inside `#decrementTokenOwner` which needs to happen first

contracts/core/extensions/CoreAccounting.sol
hydrazines: Just changed it back, I see the `require` in `decrementTokenOwner` that needs to happen first. I originally changed it to mirror deposit where the transfer happens first and then the increment.

contracts/core/extensions/CoreAccounting.sol
```diff
@@ -266,4 +266,41 @@ contract CoreAccounting is
             );
         }
     }
+
+    /**
+     * Internal function that withdraws multiple tokens from the vault.
+     * Quantities should be in the order of the addresses of the tokens being withdrawn.
+     *
+     * @param  _from              Address to decredit for withdrawls
```
justinkchen: `withdrawals`

contracts/core/extensions/CoreAccounting.sol
justinkchen: When is it supposed to be `_from` and `_to` being different?

contracts/core/extensions/CoreAccounting.sol
hydrazines: +1

contracts/core/extensions/CoreAccounting.sol
hydrazines: I don't think there's a use case where they're different right now, thoughts? Added it here to mirror how it's done during `depositInternal`

PR #268, opened 2018-11-06
Move transfer quantity check to root internal function in vault

hydrazines:


contracts/core/Vault.sol
```diff
@@ -195,9 +195,13 @@ contract Vault is
 
         for (uint256 i = 0; i < _tokens.length; i++) {
             uint256 quantity = _quantities[i];
-            if (quantity > 0) {
-                transferBalanceInternal(_tokens[i], _from, _to, quantity);
-            }
+
+            transferBalanceInternal(
+                _tokens[i],
```
justinkchen: we use `from`, `to`, `token` in your other PR as the order, should we do that here too?

hydrazines: Hmm, I'm for this if no one has any objections. If we reorder it to `from`, `to`, `token`, we should do that for all the function arguments in `Vault`.

And if that's the case, might also make sense to change the order of `getOwnerBalance`'s arguments to `owner`, `token` since intuitively you're checking a user's token balance in that order (though in implementation it's the other way around).

hydrazines: I'll save that for another PR in any case. Okay to merge these changes in for now?

asoong: It should be `tokens, from, to, quantity`

PR #269, opened 2018-11-07
Make Invariant Labels Constant Strings Variables

aunyks:
This improves the readability and maintainability of error handling in contracts.


bweick:
Thanks again for contributing! We originally had our contracts set up this way, however it added a lot of cost to deployment since it bloats the byte size of the contracts. Run time costs are fairly similar (in line was actually cheaper) so we decided to move all our error messaging in line since it was much cheaper and we felt error messaging should be unique for each `require` to make for a better dev experience.

aunyks:
Cool! Thanks for the clarification.

PR #270, opened 2018-11-07
Make IERC20 comply with ERC20 standard and edit ERC20Wrapper tests

aunyks:
[The standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md#transfer) defines that `transfer` and `transferFrom` must return a boolean value indicating the operations' success. Assuming that the two follow the standard, we can use their return values in the invariant `require` statements in `ERC20Wrapper`.

It should be acknowledged that a lot of "ERC20-like" tokens don't actually follow the standard due to not returning a value, but that behavior should be covered in a different manner than was previously defined. [This](https://github.com/aragon/aragon-apps/issues/562) is an issue that Aragon has related to this fact.

### DIFF
--- a/contracts/lib/ERC20Wrapper.sol
+++ b/contracts/lib/ERC20Wrapper.sol
@@ -30,6 +30,12 @@ import { IERC20 } from "./IERC20.sol";
  */
 library ERC20Wrapper {
 
+    // ============ Internal Constants ============
+
+    string constant internal  UNSUCCESSFUL_ERC20_TRANSFER = "XFER_BAD_ERC20_RETURN_VALUE";
+    string constant internal UNSUCCESSFUL_ERC2O_TRANSFERFROM = "XFERFROM_BAD_ERC20_RETURN_VALUE";
+    string constant internal UNSUCCESSFUL_ERC20_APPROVAL = "APPROVE_BAD_ERC20_RETURN_VALUE";
+
     // ============ Internal Functions ============
 
     /**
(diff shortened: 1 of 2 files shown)


bweick:
Hey @aunyks thanks for contributing! So for this PR we check that the token being transferred returns either true or we also accept transfers that don't return anything. The `checkSuccess` function checks the last returned value (which would be what the transfer function returns) to make sure it returns a valid value. Our implementation is very similar to the one linked to in your PR.

aunyks:
@bweick Thanks for getting back! I think conforming `IERC20` to the standard is necessary for standardized interoperability with compliant tokens. And, I still think that using the standard's intended return types provide a significant boost in readability, as I don't find `checkSuccess` necessary if the functions already return true or a "falsey" value.

bweick:
Well let's say it doesn't return anything (I believe OmiseGo's tokens don't return a value) then this function will revert so we would never be able to transfer OMG tokens. We have to define some standard that we will accept and we felt that obviously `true` and no value are acceptable return values.

aunyks:
Ugh, that's annoying lol. In that case, I still think the update to `IERC20` should remain, as the standard is finalized. With that being said, however, I can revert `ERC20Wrapper` to use `checkSuccess` again to account for tokens that don't actually return values.

bweick:
Sorry for getting back late on this. So the problem with adding the boolean to IERC20 is that when we go to interact with OMG's contract the interfaces won't match up because it's expecting a `transfer` function that returns a boolean but it in fact doesn't return anything. So in order to accommodate for weird return values we can't put any restrictions on the return portion of the interface, otherwise the transfer function won't be found and it will revert! Does that make sense?

aunyks:
I see what you're saying.

PR #271, opened 2018-11-07
Upgrade to open-zeppelin-solidity

felix2feng:


contracts/core/SetToken.sol
```diff
@@ -184,17 +177,7 @@ contract SetToken is
         // Check that function caller is Core
         require(msg.sender == ISetFactory(factory).core(), "ONLY_CORE_CAN_BURN_SET");
 
-        // Require user has tokens to burn
```
hydrazines: We still want this check right?

felix2feng: It's actually checked already in the `_burn` function of ERC20.sol

contracts/core/RebalancingSetToken.sol
```diff
@@ -532,16 +522,9 @@ contract RebalancingSetToken is
...
 
         // Require user has tokens to burn
-        require(balances[_from] >= _quantity, "NOT_ENOUGH_TOKENS_TO_BURN");
-
-        // Update token balance of user
-        balances[_from] = balances[_from].sub(_quantity);
-
-        // Update total supply of Set Token
-        totalSupply_ = totalSupply_.sub(_quantity);
+        require(balanceOf(_from) >= _quantity, "NOT_ENOUGH_TOKENS_TO_BURN");
```
hydrazines: Should we get rid of this check here then?

contracts/core/SetToken.sol
hydrazines: Oh gotcha, see my other comment

contracts/core/RebalancingSetToken.sol
felix2feng: Yes, we can do that :)

justinkchen:
Need to update `.solcover.js` too i think

contracts/core/RebalancingSetToken.sol
```diff
@@ -492,24 +492,14 @@ contract RebalancingSetToken is
         (managerFee, protocolFee) = calculateFeeSplit(totalFees);
 
         // Update token balance of the manager
-        balances[_issuer] = balances[_issuer].add(issuerTotal);
-        balances[manager] = balances[manager].add(managerFee);
+        _mint(_issuer, issuerTotal);
```
asoong: That's very convenient

asoong: @bweick

contracts/mocks/tokens/BadTokenMock.sol
```diff
@@ -31,7 +30,7 @@ contract BadTokenMock is StandardToken {
   */
   function transfer(address _to, uint256 _value) public returns (bool) {
     require(_to != address(0));
-    require(_value <= balances[msg.sender]);
+    require(_value <= balanceOf(msg.sender));
```
asoong: I wonder if this adds to the call stack by making it a function that just accesses the mapping

felix2feng: I assume it does

PR #273, opened 2018-11-08
Upgrade 0.4.25 compiler version and related configs

justinkchen:
Also pulls in AddressArrayUtils from cryptofin-solidity into our `lib` folder


contracts/lib/AddressArrayUtils.sol
```diff
@@ -0,0 +1,346 @@
+pragma solidity 0.4.25;
```
felix2feng: Can we add a link to the cryptofin repo? Also could we pull in their test files as well (we may put it in a new test folder)

asoong: +1

justinkchen: what do you mean a link to the cryptofin repo?

PR #274, opened 2018-11-08
Remove component struct

asoong:
Struct that doesn't serve any purpose: unnecessary gas costs incurred during `rebalance`,`issue`,`redeem`


contracts/core/SetToken.sol
```diff
@@ -127,10 +121,8 @@ contract SetToken is
             isComponent[currentComponent] = true;
 
             // Add component data to components struct array
```
hydrazines: Nit: Update comment

PR #278, opened 2018-11-09
Move requires into root issueInternal function in CoreIssuance and fix comments/whitespace

hydrazines:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -36,43 +36,43 @@ contract CoreInternal is
 
     /* ============ Events ============ */
 
-    // Logs registration of new exchange conforming to IExchangeWrapper
+    // Logs a factory registration change; factory must conform to ISetFactory
+    event FactoryRegistrationChanged(
```
hydrazines: Just rearranged to reflect the order in which they're used later.

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -226,18 +215,29 @@ contract CoreIssuance is
     )
         internal
     {
+        // Verify Set was created by Core and is enabled
+        require(
```
hydrazines: The only material change in this PR. Moved checks into internal function whereas before, these checks weren't happening in `CoreIssuanceOrder`.

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -316,34 +316,35 @@ contract CoreIssuance is
...
 
         // Fetch Set token properties
         uint256 naturalUnit = setToken.naturalUnit();
         address[] memory components = setToken.getComponents();
-        uint[] memory units = setToken.getUnits();
+        uint256[] memory units = setToken.getUnits();
 
         // Transfer the underlying tokens to the corresponding token balances
         for (uint256 i = 0; i < components.length; i++) {
-            address currentComponent = components[i];
```
hydrazines: We weren't doing this in other similar functions so I got rid of it.

PR #283, opened 2018-11-13
RebalancingSetToken: Updates to naming + add event

felix2feng:
Completes this [card](https://trello.com/c/hKaNqaGp/719-audit-rebalancingsettokensol-sundry-items)


contracts/core/RebalancingSetToken.sol
```diff
@@ -106,6 +106,11 @@ contract RebalancingSetToken is
         address newSet
     );
 
+    event BidPlaced(
```
bweick: I wonder if it makes sense to put this in CoreRebalanceAuction? Would mean less system wide code dup and would actually be thrown when everything is reconciled with the bid (wouldn't throw in case of revert but also would save on the gas of calling it if it reverts later).

PR #285, opened 2018-11-15
Add Operational States

felix2feng:
Note that for withdraw/redeem functions, there is no modifier as it is assumed that those functions will always be functional.

### DIFF
--- a/contracts/core/extensions/CoreAccounting.sol
+++ b/contracts/core/extensions/CoreAccounting.sol
@@ -16,6 +16,7 @@
 
 pragma solidity 0.4.25;
 
+import { CoreOperationState } from "./CoreOperationState.sol";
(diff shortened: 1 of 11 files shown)


contracts/core/extensions/CoreAccounting.sol
```diff
@@ -16,6 +16,7 @@
 
 pragma solidity 0.4.25;
 
+import { CoreOperationState } from "./CoreOperationState.sol";
```
asoong: relative imports after absolute

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -16,6 +16,7 @@
 
 pragma solidity 0.4.25;
 
+import { CoreOperationState } from "./CoreOperationState.sol";
```
asoong: same

PR #286, opened 2018-11-15
CoreInternal: Convert register functions to add/remove

felix2feng:


contracts/core/extensions/CoreInternal.sol
```diff
@@ -111,68 +130,123 @@ contract CoreInternal is
     {
         state.exchanges[_exchangeId] = _exchange;
 
-        emit ExchangeRegistered(
+        emit ExchangeRegistrationChanged(
```
asoong: I would make this a separate event

PR #287, opened 2018-11-16
First pass at updating contracts to modular format. Further testing r…

bweick:
…equired.


contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -107,7 +145,7 @@ contract CoreIssuanceOrder is
         );
 
         // Verify signature is authentic
-        ISignatureValidator(state.signatureValidator).validateSignature(
+        ISignatureValidator(ICore(core).signatureValidator()).validateSignature(
```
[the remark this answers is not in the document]
bweick: I think the contract tracking should still all run through Core which is why I kept it there.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -40,15 +38,29 @@ import { OrderLibrary } from "../lib/OrderLibrary.sol";
  * The Core Issuance Order extension houses all functions related to the filling and
  * canceling of issuance orders.
  */
-contract CoreIssuanceOrder is
-    ICoreIssuance,
-    ICoreAccounting,
-    CoreState,
+contract IssuanceOrderModule is
```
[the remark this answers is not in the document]
bweick: I prefer the latter cause it rolls off the tongue better. It's also it's own contract so don't feel like it falls under the same category as our core extension type contracts when we prefix with Core.

contracts/core/extensions/CoreModuleInteraction.sol
```diff
@@ -0,0 +1,158 @@
...
+     * @param  _to              Address to credit for deposit
+     * @param  _quantity        Amount of tokens to deposit
+     */
+    function depositModule(
+        address _token,
+        address _from,
+        address _to,
+        uint256 _quantity
+    )
+        external
```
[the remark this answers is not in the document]
bweick: Yup added in newest commit

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -171,6 +174,20 @@ contract KyberNetworkWrapper {
         );
     }
 
+    /**
+     * Set the CoreIsssuanceModule so that only it can call exchange.
+     *
+     * @param  _newIssuanceOrderModule          Address of deployed issuanceOrderModule contract
+     */    
+    function setIssuanceOrderModule(
```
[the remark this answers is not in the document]
bweick: I think as little redeployment of contracts as possible is best. Not necessarily because its cheaper but because we should try to make the system as static as possible, people in theory will be watching this contracts quite a bit and moving them around with a redeploy feels less scalable.

contracts/core/RebalancingSetTokenFactory.sol
```diff
@@ -156,6 +165,20 @@ contract RebalancingSetTokenFactory {
...
+    /**
+     * Set the RebalanceAuctionModule so that only it can call placeBid on rebalancingSetToken.
+     *
+     * @param  _newRebalanceAuctionModule          Address of deployed rebalanceAuctionModule contract
+     */    
+    function setRebalanceAuctionModule(
+        address _newRebalanceAuctionModule
+    )
+        external
+        onlyOwner
```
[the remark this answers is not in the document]
bweick: Yeah we should...another PR though once yours is merged

contracts/core/extensions/CoreModuleInteraction.sol
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Exposes internal function that deposits a quantity of tokens to the vault and attributes
+     * the tokens respectively, to system modules.
+     *
+     * @param  _token           Address of token being deposited
+     * @param  _from            Address to transfer tokens from
+     * @param  _to              Address to credit for deposit
+     * @param  _quantity        Amount of tokens to deposit
+     */
+    function depositModule(
```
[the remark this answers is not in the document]
bweick: Yeah I think this might be a better approach

contracts/core/interfaces/ICore.sol
```diff
@@ -268,55 +291,85 @@ interface ICore {
...
      *
-     * @param  _rebalancingSetToken        The address of the rebalancing token being bid on
-     * @param  _quantity                   The number of currentSets to rebalance
+     * @param  _owner        Address to issue tokens to
+     * @param  _set          Address of the Set to issue
+     * @param  _quantity     Number of tokens to issue
      */
-    function bid(
-        address _rebalancingSetToken,
+    function issueModule(
```
felix2feng: If we don't upgrade Core, we might want to expose other functions like `redeemInternal` as well

contracts/core/lib/CoreState.sol
```diff
@@ -188,6 +185,22 @@ contract CoreState {
         return state.validFactories[_factory];
     }
 
+    /**
+     * Return boolean indicating if address is valid module.
+     *
+     * @param  _module        Factory address
+     * @return bool           Boolean indicating if enabled factory
+     */
+    function validModules(
```
felix2feng: validModule vs validModules?

contracts/core/extensions/CoreInternal.sol
```diff
@@ -98,6 +98,36 @@ contract CoreInternal is
...
+     *
+     * @param _module     Address of the module
+     */
+    function addModule(
+        address _module
+    )
+        external
+        onlyOwner
+    {
+        state.validModules[_module] = true;
```
felix2feng: We should emit an event, so we can keep track of history

contracts/core/extensions/CoreInternal.sol
```diff
@@ -98,6 +98,36 @@ contract CoreInternal is
...
+     *
+     * @param _module   Enumeration of module within the mapping
+     */
+    function removeModule(
+        address _module
+    )
+        external
+        onlyOwner
+    {
+        state.validModules[_module] = false;
```
felix2feng: Same re: event. But we can do it later.

contracts/core/RebalancingSetToken.sol
```diff
@@ -411,9 +411,9 @@ contract RebalancingSetToken is
         external
         returns (address[], uint256[], uint256[])
     {
-        // Make sure sender is Core
+        // Make sure sender is rebalanceAuctionModule
         require(
-            msg.sender == IRebalancingSetFactory(factory).core(),
+            msg.sender == IRebalancingSetFactory(factory).rebalanceAuctionModule(),
```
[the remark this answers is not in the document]
bweick: Main reason I didn't add specific module names to the Core is because when we add new modules we won't be able to add their specific identifier to Core. As I think about it we don't actually need to specifically identify which module is calling these functions just that a module we approve of is, so I'd be fine just doing something like `ICore(core).validModule(msg.sender)` which I think accomplishes the same thing instead of specifying the module we need.

contracts/core/interfaces/ICore.sol
bweick: +1

contracts/core/RebalancingSetToken.sol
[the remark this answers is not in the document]
felix2feng: +1

contracts/core/extensions/CoreInternal.sol
bweick: Lets jsut add that to your PR since you have a standard for how you've been doing those.

felix2feng: sure

contracts/core/lib/CoreState.sol
felix2feng: ok

contracts/core/RebalancingSetToken.sol
```diff
@@ -411,9 +411,9 @@ contract RebalancingSetToken is
         external
         returns (address[], uint256[], uint256[])
     {
-        // Make sure sender is Core
+        // Make sure sender is rebalanceAuctionModule
```
felix2feng: In actuality, we're checking that its a valid module

contracts/core/RebalancingSetTokenFactory.sol
```diff
@@ -29,7 +30,9 @@ import { LibBytes } from "../external/0x/LibBytes.sol";
  * RebalancingSetTokens deployed by the factory can only have their mint and burn functions
  * called by Core
  */
-contract RebalancingSetTokenFactory {
+contract RebalancingSetTokenFactory is
+    Ownable
```
felix2feng: Do we still need this?

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -29,7 +31,9 @@ import { LibBytes } from "../../external/0x/LibBytes.sol";
  *
  * The KyberNetworkWrapper contract wrapper to interface with KyberNetwork for reserve liquidity
  */
-contract KyberNetworkWrapper {
+contract KyberNetworkWrapper is 
+    Ownable
```
felix2feng: Same - don't think we need it.

contracts/core/exchange-wrappers/TakerWalletWrapper.sol
```diff
@@ -29,7 +31,9 @@ import { LibBytes } from "../../external/0x/LibBytes.sol";
  *
  * The TakerWalletWrapper contract wrapper to transfer tokens directly from order taker
  */
-contract TakerWalletWrapper {
+contract TakerWalletWrapper is 
+    Ownable
```
felix2feng: Same here

contracts/core/RebalancingSetToken.sol
```diff
@@ -411,9 +411,9 @@ contract RebalancingSetToken is
...
         require(
-            msg.sender == IRebalancingSetFactory(factory).core(),
+            ICore(IRebalancingSetFactory(factory).core()).validModules(msg.sender),
             "RebalancingSetToken.placeBid: Sender must be core"
```
felix2feng: Update me

contracts/core/exchange-wrappers/ZeroExExchangeWrapper.sol
```diff
@@ -34,7 +36,9 @@ import { ZeroExOrderDataHandler as OrderHandler } from "./lib/ZeroExOrderDataHan
  *
  * The ZeroExExchangeWrapper contract wrapper to interface with 0x V2
  */
-contract ZeroExExchangeWrapper {
+contract ZeroExExchangeWrapper is
+    Ownable
```
felix2feng: Same

PR #288, opened 2018-11-16
Time Lock Upgradability

felix2feng:
- Moved TimeLockUpgrade into its own Library
- TimeLockUpgrade is used in `CoreInternal` and `Authorizable`


test/contracts/core/extensions/coreTimeLockUpgrade.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+          { from: subjectCaller },
+        );
+      });
+
+      it('should register the timeLockedUpgrade with the transaction timestamp', async () => {
+        const txHash = await subject();
+        const { blockHash, input } = await web3.eth.getTransaction(txHash);
+        const { timestamp } = await web3.eth.getBlock(blockHash);
+
+        const upgradeHash = web3.utils.soliditySha3(input);
```
felix2feng: Really cool that web3 1.0 now has a pack and hash function

test/contracts/core/extensions/coreTimeLockUpgrade.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      beforeEach(async () => {
+        await coreTimeLockUpgradeMock.setTimeLockPeriod.sendTransactionAsync(
+          ONE,
+          { from: subjectCaller },
+        );
+      });
+
+      it('should register the timeLockedUpgrade with the transaction timestamp', async () => {
+        const txHash = await subject();
+        const { blockHash, input } = await web3.eth.getTransaction(txHash);
```
felix2feng: `input` = calldata

contracts/core/extensions/CoreInternal.sol
```diff
@@ -240,7 +246,26 @@ contract CoreInternal is
     )
         external
         onlyOwner
+        timeLockUpgrade
     {
         state.signatureValidator = _signatureValidator;
     }
+
```
bweick: Why not put this in a constructor instead of having a function that only gets called once?

felix2feng: It kills 2 birds with 1 stone, which is 1) adding an initial period where the timelock doesn't apply (not setting it has that effect) and 2) allows adding that timelock

bweick: I see. That makes sense

PR #293, opened 2018-11-19
Only check sig if order hasn't been filled before. Don't transfer 0 a…

bweick:
…mount to taker. Formatting.

### DIFF
--- a/contracts/core/modules/IssuanceOrderModule.sol
+++ b/contracts/core/modules/IssuanceOrderModule.sol
@@ -126,7 +126,7 @@ contract IssuanceOrderModule is
      */
     function fillOrder(
         address[5] _addresses,
-        uint[6] _values,
+        uint256[6] _values,
         address[] _requiredComponents,
         uint256[] _requiredComponentAmounts,
         uint256 _fillQuantity,
@@ -145,11 +145,13 @@ contract IssuanceOrderModule is
         );
 
         // Verify signature is authentic
-        ISignatureValidator(ICore(core).signatureValidator()).validateSignature(
-            order.orderHash,
-            order.makerAddress,
-            _signature
-        );
+        if (orderFills[order.orderHash] == 0) {
+            ISignatureValidator(ICore(core).signatureValidator()).validateSignature(
+                order.orderHash,
+                order.makerAddress,
+                _signature
+            );            
+        }
 
         // Verify order is valid and return amount to be filled
         validateOrder(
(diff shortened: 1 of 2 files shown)


contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -145,11 +145,13 @@ contract IssuanceOrderModule is
         );
 
         // Verify signature is authentic
-        ISignatureValidator(ICore(core).signatureValidator()).validateSignature(
-            order.orderHash,
-            order.makerAddress,
-            _signature
-        );
+        if (orderFills[order.orderHash] == 0) {
```
felix2feng: Can we add a comment that we add the check as an optimization?

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -513,13 +515,15 @@ contract IssuanceOrderModule is
...
     {
-        // Send left over maker token balance to taker
-        ITransferProxy(transferProxy).transfer(
-            _order.makerToken,
-            _requiredMakerTokenAmount.sub(_makerTokenUsed), // Required less used is amount sent to taker
-            _order.makerAddress,
-            msg.sender
-        );
+        // Send left over maker token balance to taker, if greater than 0
+        if (_requiredMakerTokenAmount.sub(_makerTokenUsed) > 0) {
```
felix2feng: We could put `_requiredMakerTokenAmount.sub(_makerTokenUsed)` into a variable since its being used twice.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -562,30 +567,33 @@ contract IssuanceOrderModule is
...
             transferProxyInstance.transfer(
                 _order.relayerToken,
                 makerFee,
                 _order.makerAddress,
                 _order.relayerAddress
             );
         }
-        if (takerFee > 0) {
-            //Send taker fees to relayer
+        if (_order.takerRelayerFee > 0) {
```
felix2feng: Add a return / space after if clause

contracts/core/modules/IssuanceOrderModule.sol
asoong: +1

test/contracts/core/modules/issuanceOrderModule.spec.ts
```diff
@@ -508,6 +532,32 @@ contract('IssuanceOrderModule', accounts => {
...
+        ABIDecoder.addABI(StandardTokenMock.abi);
+        zeroExOrderTakerAssetAmount = ether(5);
+      });
+
+      after(async () => {
+        ABIDecoder.removeABI(StandardTokenMock.abi);
+        zeroExOrderTakerAssetAmount = undefined;
+      });
+
+      it('does not execute a transfer of the relayer fees for 0 amount', async () => {
```
asoong: Nice way of checking.

PR #296, opened 2018-11-21
Enabling partial fills for issuance orders.

bweick:


test/contracts/core/exchange-wrappers/zeroExExchangeWrapper.spec.ts
```diff
@@ -438,20 +448,17 @@ contract('ZeroExExchangeWrapper', accounts => {
...
+        subjectValues,
+        subjectOrderData,
+        { from: subjectCaller, gas: DEFAULT_GAS },
+      );
+    }
 
       it('should correctly return the fill Results', async () => {
         const [tokens, fillAmounts] = await subject();
-
+        console.log(tokens, fillAmounts);
```
felix2feng: remove this

test/contracts/core/exchange-wrappers/zeroExExchangeWrapper.spec.ts
```diff
@@ -400,6 +408,8 @@ contract('ZeroExExchangeWrapper', accounts => {
           zeroExOrderFillAmount
         );
         subjectOrderData = SetTestUtils.concatBytes([zeroExExchangeWrapperOrder, secondZeroExExchangeWrapperOrder]);
+        console.log(subjectOrderData.length);
```
felix2feng: remove

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+        view
+    {
+        // Verify maker token used is less than amount allocated that user signed
+        require(
+            _makerTokenAmountUsed <= _requiredMakerTokenAmount,
+            "IssuanceOrderModule.settleOrder: Maker token used exceeds allotted limit"
         );
 
+        // Check that maker's component tokens in Vault have been incremented correctly
+        for (uint16 i = 0; i < _order.requiredComponents.length; i++) {
```
felix2feng: uint256

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+        OrderLibrary.IssuanceOrder _order,
+        uint256[] _requiredBalances,
+        uint256 _requiredMakerTokenAmount,
+        uint256 _makerTokenAmountUsed
+    )
+        private
+        view
+    {
+        // Verify maker token used is less than amount allocated that user signed
+        require(
```
felix2feng: https://blog.codinghorror.com/curlys-law-do-one-thing/

The name of this function is `checkRequiedTokenBalances` - thus I would expect it to just check the balances of the required tokens. I'd move this require up on the level of the `settleOrder` function.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+     * Check exchange orders acquire correct amount of tokens and taker doesn't over use
+     * the issuance order maker's tokens
+     *
+     * @param  _order                       IssuanceOrder object containing order params
+     * @param  _requiredBalances            Array of required balances for each component
+                                            after exchange orders are executed
+     * @param  _requiredMakerTokenAmount    Max amount of maker token used to source tokens
+     * @param  _makerTokenAmountUsed        Amount of maker token used to source tokens
+     */
+    function checkRequiredTokenBalances(
```
felix2feng: https://softwareengineering.stackexchange.com/questions/166884/should-i-extract-specific-functionality-into-a-function-and-why

The thing stand out to me in the above stackoverflow article is "choose good and precise names for your functions". It's generally a really good practice to take an extra second to think about how you will name your function and what the components of the function will be.

Also - We generally use the terms `assert` or `validate`. Perhaps a better name that reflects what this function does is `assertTokenAcquisitionQuantities`.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+            // Read the order body based on order data length info in header plus the length of the header (128)
+            uint256 exchangeDataLength = header.orderDataBytesLength.add(128);
+            bytes memory bodyData = LibBytes.slice(
+                _orderData,
+                scannedBytes.add(128),
+                scannedBytes.add(exchangeDataLength)
+            );
+
+            // Calculate amount of makerToken actually needed
+            uint256 neededMakerTokenAmount = OrderLibrary.getPartialAmount(
```
felix2feng: What if `_fillQuantity == `_attemptedFillQuantity`? Will we still need to calculate partialAmount each time as well as in the exchange wrappers?

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+            ITransferProxy(transferProxy).transfer(
+                _makerTokenAddress,
+                neededMakerTokenAmount,
+                _makerAddress,
+                exchangeWrapper
+            );
+
+            // Call Exchange
+            callExchange(
+                [_makerAddress, msg.sender, _makerTokenAddress],
```
felix2feng: Would the `taker` ever not be `msg.sender`?

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -139,33 +141,27 @@ contract KyberNetworkWrapper {
 
         // Ensure the issuance order maker token is allowed to be transferred by KyberNetworkProxy as the source token
         ERC20.ensureAllowance(
-            _makerToken,
+            _addresses[2],
```
felix2feng: For clarity, I would instantiating these array addresses/uints as variables before usage. As I read the code, it is jarring to go back and forth to the comments above to understand what is going on.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -236,109 +235,32 @@ contract IssuanceOrderModule is
...
-        private
+        public
         view
+        returns (uint256)
     {
         // Declare set interface variable
         ISetToken set = ISetToken(_order.setAddress);
 
+        // Determine amount to execute
+        uint256 closedOrderAmount = orderFills[_order.orderHash].add(orderCancels[_order.orderHash]);
```
felix2feng: What is the purpose of `validateOrder`? As I read the function name, I understand it to "validate an order" - which is making assertions that either validate or invalidate the order. I don't expect it to do anything else.

Checking how much of a Set is executable is separate from validating an order, so it would make sense to put into another function. I also wouldn't expect (as a user/developer) to receive the executable amount from calling validateOrder. I would expect a boolean, a revert/no revert.

Adding this additional functionality makes this function do more than validate an order (doing more than a single thing) and it is logically misplaced.

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+    function callExchange(
+        address[3] _addresses,
+        uint256[4] _values,
+        address _exchange,
+        bytes _bodyData
+    )
+        private
+    {
+        // Call Exchange
+        address[] memory componentFillTokens = new address[](_values[1]);
```
felix2feng: Same comment re: illegibility of `_values[1]`

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -405,48 +327,194 @@ contract IssuanceOrderModule is
...
+            );
+            require(
+                currentBal >= _requiredBalances[i],
+                "IssuanceOrderModule.settleOrder: Insufficient component tokens acquired"
+            );
+        }        
+    }
+
+    /**
+     * Check exchange orders acquire correct amount of tokens and taker doesn't over use
```
felix2feng: Update comment

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -460,33 +528,58 @@ contract IssuanceOrderModule is
...
+     *
+     * @param  _order                   IssuanceOrder object containing order params
+     * @param  _fillQuantity            Quantity of Set to be filled
+     * @param  _attemptedFillQuantity   Quantity of Set taker attempted to fill
+     * @param  _orderData               Bytestring encoding all exchange order data
+     */
+    function settleOrder(
+        OrderLibrary.IssuanceOrder _order,
+        uint256 _fillQuantity,
+        uint256 _attemptedFillQuantity,
```
felix2feng: It is hard to understand the delineation between `_fillQuantity` and `attemptedFillQuantity`. I don't know how we could name this better off the top of my head... but it needs to be more clear

contracts/core/modules/IssuanceOrderModule.sol
bweick: Unfortunately yeah. It's not ideal, unless we want to add a lot of `if...else...` statements, which we can do.

contracts/core/modules/IssuanceOrderModule.sol
bweick: Well we can change the function name if you think it's misleading, I do think that part of validating an order is knowing how much you're gonna get back. But I still don't see why this needs to be split into separate functions. It's adding unnecessary gas, for little clarity.

contracts/core/modules/IssuanceOrderModule.sol
bweick: `_actualFillQuantity`?

contracts/core/modules/IssuanceOrderModule.sol
bweick: I really do think there has to be a difference in the way we approach solidity development vs traditional. I get the argument that clarity is very important for security reasons. At the same time, there's a balance between that and writing efficient code in this instance. We have to be ok with the idea that people are going to need to grasp compound concepts. I don't think it's practical at this point to separate out all logic into it's own function, there's some middle ground here.

felix2feng: When you are talking about efficient code, I'm assuming you mean gas-efficient code. In that case, I believe that writing well-abstracted and understandable code allows us to achieve that goal. When writing efficient code or optimizing, the first place to start is removing unnecessary or redundant SSTORE opcodes. It becomes extremely easier to identify where we could make these fixes when the code is easier to reason about. 

https://github.com/ethereum/pyethereum/blob/develop/ethereum/opcodes.py
^ Opcodes and cost. If you look at the costs of operations, `SSTORE`  (5,000 - 20k gas) is by far the most expensive. `EXTCODECOPY` (now 700 for external contract loading) and `JUMP` (8 gas) are extremely low cost. When the costs are so low, I prefer clarity over gas costs.

Obviously, we don't want to over-abstract - which is when the code itself is clearer than the function. As someone looking at our existing code, it is not easy to quickly understand what is happening. The side effects of large functions it is that we hit stack too deep errors (as we're cramming too much into a function).

Re: `assert` / `validate`, it seems we prefer validate (e.g. `validateSignature``validateOrder`). The only deviance is `checkSuccess` in ERC20Wrapper. Let's go with `validate` going forward.

P.S. I believe there are numerous places in `IssuanceOrderModule` and the `ExchangeWrappers` where we can save gas through duplicate operations.

contracts/core/modules/IssuanceOrderModule.sol
felix2feng: As I think about it more, `validateOrder` should only take the `OrderLibrary.IssuanceOrder` parameter. Whether the order is fillable with the `executeQuantity` and how much it can be filled is a different operation. Thus, the function would truly be validating whether an order is valid.

Then we can write a different function called `validateOrderExecutable` that takes in the `orderHash` and `executeQuantity` which can check whether the order is executable as well as return the executable amount.

asoong: This should not be in `validateOrder`. Is there a stack size issue here?

contracts/core/modules/IssuanceOrderModule.sol
felix2feng: Should we even pass msg.sender into callExchange vs passing `msg.sender` in call exchange to the exchange wrapper?

contracts/core/modules/IssuanceOrderModule.sol
felix2feng: I don't have an answer.. but there might be an architectural / structural change that we can make to reduce the duplication of logic in calculating the partialAmount.

asoong: Let's get this merged first, and then we can see what the invocation cost differences are with bypassing the calculations. Add a trello for this

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -139,33 +141,35 @@ contract KyberNetworkWrapper {
...
         );
 
-        address[] memory componentTokensReceived = new address[](_tradeCount);
-        uint256[] memory componentTokensAmounts = new uint256[](_tradeCount);
+        OrderLibrary.FractionFilled memory fractionFilled = OrderLibrary.FractionFilled({
+            filled: _values[2],
+            attempted: _values[3]
+        });
+
+        uint256 orderCount = _values[1];
```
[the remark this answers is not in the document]
felix2feng: Nope- reverting to tradesCount

contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -139,33 +141,35 @@ contract KyberNetworkWrapper {
...
+            _addresses[2], // makerToken
             address(this),
             kyberNetworkProxy,
-            _makerAssetAmount
+            _values[0] // makerAssetAmount
         );
 
-        address[] memory componentTokensReceived = new address[](_tradeCount);
-        uint256[] memory componentTokensAmounts = new uint256[](_tradeCount);
+        OrderLibrary.FractionFilled memory fractionFilled = OrderLibrary.FractionFilled({
```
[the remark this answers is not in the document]
felix2feng: Let's reserve this change for the next PR

contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -606,4 +557,105 @@ contract IssuanceOrderModule is
...
+        uint256 _executeQuantity
+    )
+        internal
+        view
+        returns (uint256)
+    {
+        uint256 closedOrderAmount = orderFills[_order.orderHash].add(orderCancels[_order.orderHash]);
+        uint256 openOrderAmount = _order.quantity.sub(closedOrderAmount);
+        
+        return openOrderAmount.min(_executeQuantity);
```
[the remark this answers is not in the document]
felix2feng: +1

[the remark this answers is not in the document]
asoong: Isn't 0 mod n still 0? that would always be true.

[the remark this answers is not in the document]
felix2feng: Hm true

PR #297, opened 2018-11-22
RebalancingSetToken code organization and touch ups

felix2feng:
- Non functionality changing code touch ups

### DIFF
--- a/contracts/core/RebalancingSetToken.sol
+++ b/contracts/core/RebalancingSetToken.sol
@@ -290,24 +290,11 @@ contract RebalancingSetToken is
             "RebalancingSetToken.rebalance: Proposal period not elapsed"
         );
 
-        // Create token arrays needed for auction
-        auctionSetUp();
-
-        // Get currentSet natural unit
-        uint256 currentSetNaturalUnit = ISetToken(currentSet).naturalUnit();
-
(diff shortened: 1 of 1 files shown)


contracts/core/RebalancingSetToken.sol
```diff
@@ -747,12 +713,18 @@ contract RebalancingSetToken is
         // Create combined token Array
         address[] memory oldComponents = currentSetInstance.getComponents();
         address[] memory newComponents = nextSetInstance.getComponents();
+
```
asoong: nit: I prefer to not have the space when newComponents was invoked so that it can be called in the next line. Also makes the comment orphaned

contracts/core/RebalancingSetToken.sol
```diff
@@ -737,7 +703,7 @@ contract RebalancingSetToken is
      * Calcualate unit difference between both sets relative to the largest natural
      * unit of the two sets.
      */
-    function auctionSetUp()
+    function setUpAuctionDataStructures()
```
asoong: setupAuctionParameters maybe? dataStructures is goofy

bweick: +1 since it does more than just set up the arrays it also defines the minimumBid too.

PR #300, opened 2018-11-27
Issuance Order / ExchangeWrappers: Refactor exchange data into structs

felix2feng:
- Thank you @justinkchen for helping with the contract wrapper changes to allow complex structs to be passed in.


contracts/core/exchange-wrappers/KyberNetworkWrapper.sol
```diff
@@ -120,18 +121,16 @@ contract KyberNetworkWrapper {
      * fillQuantity                     Quantity of Set to be filled
      * attemptedfillQuantity            Quantity of Set taker attempted to fill
      *
-     * @param  _addresses               [maker, --, makerToken]
```
asoong: Remove the rest of the parameters as well? They should refer to the struct

felix2feng: Hmm? What do you mean

PR #301, opened 2018-11-27
Update fixes

asoong:


contracts/core/RebalancingSetToken.sol
```diff
@@ -522,9 +522,9 @@ contract RebalancingSetToken is
             "RebalancingSetToken.mint: Cannot mint during Rebalance"
         );
 
-        uint256 issuerTotal;
-        uint256 managerFee;
-        uint256 protocolFee;
+        uint256 issuerTotal = 0;
```
asoong: Is there a way to initialize these all at once

PR #302, opened 2018-11-27
Brian/update rebalancing for new price curves

bweick:


bweick:
Have the price bound checking on the auction library now, because in the future we will want that to happen for any oracle based system.

contracts/core/RebalancingSetToken.sol
```diff
@@ -231,16 +231,17 @@ contract RebalancingSetToken is
             "RebalancingSetToken.propose: Invalid or disabled PriceLibrary address"
         );
 
-        // Assert price divisor is non-zero, ensuring a positive slope
+        // Check that time to pivot is within in 6 hours to 3 days
         require(
-            _auctionPriceDivisor > 0,
-            "RebalancingSetToken.propose: Price divisor must be positive"
+            _auctionTimeToPivot > 21600 && _auctionTimeToPivot < 259200,
```
felix2feng: In a following PR, let's allow these params to be passed into the constructor of the RebalancingSetToken and the RebalancingSetTokenFactory

contracts/core/lib/auction-price-libraries/IAuctionPriceCurve.sol
```diff
@@ -25,19 +25,47 @@ pragma solidity 0.4.25;
...
+        returns (uint256);
+
+    /*
+     * Validate any auction parameters that have library-specific restrictions
+     *
+     * @param  _auctionTimeToPivot        Time until auction reaches pivot point
+     * @param  _auctionStartPrice         The price to start the auction at
+     * @param  _auctionPivotPrice         The price at which auction curve changes from linear to exponential
+     */
+
```
felix2feng: extra space

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -28,27 +28,66 @@ import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
...
+    function validateAuctionPriceParameters(
+        uint256,
+        uint256,
+        uint256 _auctionPivotPrice
+    )
+        external
+        view
+    {
+        // Generically, require pivot price to be between 0.5 and 5
+        require(_auctionPivotPrice > priceDenominator.div(2) && _auctionPivotPrice < priceDenominator.mul(5));
```
felix2feng: Can you put each check on its own line? I'd also prefer that constants / magic numbers are defined at the top of the file and referenced.

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -28,27 +28,66 @@ import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
...
+        public
+    {
+        // Set price to be returned by library
+        priceDenominator = _priceDenominator;
+    }
+
+    /*
+     * Validate any auction parameters that have library-specific restrictions
+     *
+     * @param  -- Unused auction time to pivot param to conform to IAuctionPriceCurve --
```
felix2feng: This `unused` construct is a bit wonky. Can we use a struct instead?

contracts/mocks/core/lib/ConstantAuctionPriceCurve.sol
```diff
@@ -26,40 +27,68 @@ pragma solidity 0.4.25;
...
     }
 
+    /*
+     * Validate any auction parameters that have library-specific restrictions
+     *
+     * @param  -- Unused auction time to pivot param to conform to IAuctionPriceCurve --
+     * @param  -- Unused auction start price to conform to IAuctionPriceCurve --
+     * @param  _auctionPivotPrice         The price at which auction curve changes from linear to exponential
+     */
+
```
felix2feng: extra space

contracts/core/RebalancingSetToken.sol
felix2feng: Also why are we validating the pivot time in the this file vs. the price curve file?

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -28,27 +28,66 @@ import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
...
         view
-        returns (uint256)
+        returns (uint256, uint256)
     {
         // Calculate how much time has elapsed since start of auction and divide by
         // timeIncrement of 30 seconds, so price changes every 30 seconds
         uint256 elapsed = block.timestamp.sub(_auctionStartTime).div(30);
 
-        return _curveCoefficient.mul(elapsed).add(_auctionStartPrice);
+        return (elapsed, priceDenominator);
```
felix2feng: Where is the math that has the linear then exponential calculations?

contracts/mocks/core/lib/ConstantAuctionPriceCurve.sol
```diff
@@ -26,40 +27,68 @@ pragma solidity 0.4.25;
...
+    function validateAuctionPriceParameters(
+        uint256,
+        uint256,
+        uint256 _auctionPivotPrice
+    )
+        external
+        view
+    {
+        // Generically, require pivot price to be between 0.5 and 5
```
felix2feng: What does it mean between 0.5 and 5? Is the price denominator supposed to be 1?

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
bweick: How so? I think a comment right above the `require` is plenty enough. No need to have any extra variables or anything.

felix2feng: https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

Also, if our plan might be to pass this in the constructor, makes sense to store in a variable.

My other comment was to do the following:

```
require(
   _auctionPivotPrice > priceDenominator.div(2) && 
  _auctionPivotPrice < priceDenominator.mul(5)
);
```

Also, let's add a require message?

contracts/core/RebalancingSetToken.sol
```diff
@@ -186,16 +186,16 @@ contract RebalancingSetToken is
...
      * @param _auctionStartPrice            The price to start the auction at
-     * @param _auctionPriceDivisor          The granularity with which the prices change
+     * @param _auctionPivotPrice            The price at which the price curve switches from linear to exponential
      */
     function propose(
         address _nextSet,
         address _auctionLibrary,
-        uint256 _curveCoefficient,
+        uint256 _auctionTimeToPivot,
         uint256 _auctionStartPrice,
```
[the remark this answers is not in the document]
felix2feng: In the Linear Auction price curve library?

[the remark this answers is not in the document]
bweick: Yeah but it's not necessary for the linear auction library because the start price is always 0

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
asoong: They should be on their own lines with proper require messages. The comment is fine

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
asoong: +1

contracts/core/RebalancingSetToken.sol
asoong: This is valid, the curve should not be aware of the requirements of a proposal in the context of a rebalancing set token.

PR #306, opened 2018-11-29
Brian/new pivot price curve

bweick:


contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -78,25 +78,50 @@ contract LinearAuctionPriceCurve {
      *
```
felix2feng: I would add an in-depth comment that describes this. I might even try to do a graphic/chart in code format

PR #307, opened 2018-11-29
Set Token and Factory Refactor to support string names and ABI Gen

felix2feng:


contracts/core/RebalancingSetTokenFactory.sol
```diff
@@ -152,15 +158,17 @@ contract RebalancingSetTokenFactory {
...
-            _name,
-            _symbol
+            parameters.name,
+            parameters.symbol
         );
     }
 
     /* ============ Private Functions ============ */
 
     function parseRebalanceSetCallData(
```
bweick: This needs a javadoc

contracts/core/RebalancingSetToken.sol
```diff
@@ -132,13 +133,13 @@ contract RebalancingSetToken is
         uint256 _rebalanceInterval,
         uint256 _entranceFee,
         uint256 _rebalanceFee,
-        bytes32 _name,
-        bytes32 _symbol
+        string _name,
```
bweick: Change the javadoc for this. Not bytes32 anymore

contracts/core/RebalancingSetToken.sol
```diff
@@ -132,13 +133,13 @@ contract RebalancingSetToken is
...
+        string _symbol
```
bweick: Same as above

PR #312, opened 2018-12-07
Update migration for 1.2.0

asoong:


truffle.js
```diff
@@ -1,4 +1,5 @@
 var HDWalletProvider = require("truffle-hdwallet-provider");
+const NonceTrackerSubprovider = require("web3-provider-engine/subproviders/nonce-tracker")
```
felix2feng: What's the nonce tracker for?

PR #313, opened 2018-12-10
[WIP] Clean up instance instantiation

felix2feng:


felix2feng:
@bweick 
Deposit without extra instantiation: 89556 gas
Deposit with instantiation each time: 89568 gas

PR #316, opened 2018-12-13
Add CoreIssuance/issueTo

felix2feng:
- Includes adding of a recipient parameter in issueInternal

### DIFF
--- a/contracts/core/extensions/CoreIssuance.sol
+++ b/contracts/core/extensions/CoreIssuance.sol
@@ -56,6 +56,31 @@ contract CoreIssuance is
     {
         issueInternal(
             msg.sender,
+            msg.sender,
+            _set,
+            _quantity
+        );
+    }
+
+    /**
+     * Exchange components for Set tokens
+     *
+     * @param  _set          Address of the Set to issue
+     * @param  _quantity     Number of tokens to issue
+     * @param  _to           Address to issue to
+     */
+    function issueTo(
+        address _set,
+        uint256 _quantity,
+        address _to
+    )
+        external
+        nonReentrant
+        whenOperational
+    {
+        issueInternal(
+            msg.sender,
+            _to,
             _set,
             _quantity
         );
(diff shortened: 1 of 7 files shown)


contracts/core/extensions/CoreIssuance.sol
```diff
@@ -56,6 +56,31 @@ contract CoreIssuance is
...
         issueInternal(
             msg.sender,
+            msg.sender,
+            _set,
+            _quantity
+        );
+    }
+
+    /**
+     * Exchange components for Set tokens
```
asoong: Can we clarify here: 1. who the components come from (caller, param, or vault), 2. who gets the Set (caller, param, or vault)

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -56,6 +56,31 @@ contract CoreIssuance is
...
+     *
+     * @param  _set          Address of the Set to issue
+     * @param  _quantity     Number of tokens to issue
+     * @param  _to           Address to issue to
+     */
+    function issueTo(
+        address _set,
+        uint256 _quantity,
+        address _to
```
bweick: Only little naggy thing I'd say is that issueInternal has the _owner and _recipient first so for consistency sake make it first here to avoid the probable refactor we decide to do later.

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -198,12 +223,14 @@ contract CoreIssuance is
     /**
      * Exchange components for Set tokens, accepting any owner
      *
-     * @param  _owner        Address to issue tokens to
+     * @param  _owner        Address to use tokens from
```
bweick: Maybe something a little more descriptive than _owner is necessary now. Like `_componentOwner` and `_setRecipient`

contracts/core/extensions/CoreIssuance.sol
felix2feng: Good call

contracts/core/extensions/CoreIssuance.sol
felix2feng: +1

contracts/core/extensions/CoreIssuance.sol
felix2feng: +1

PR #317, opened 2018-12-13
Refactor magic numbers into constants

felix2feng:
Replace with constants where I could find it


contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -30,6 +30,11 @@ import { RebalancingHelperLibrary } from "../RebalancingHelperLibrary.sol";
 contract LinearAuctionPriceCurve {
     using SafeMath for uint256;
 
+    uint256 constant public MIN_PIVOT_PRICE_DIVISOR = 2;
+    uint256 constant public MAX_PIVOT_PRICE_NUMERATOR = 5;
```
bweick: Just one thing...can we do the same for the ConstantAuctionPriceCurve. Just because that's mainly used for testing

PR #324, opened 2018-12-16
Minor gas usage improvements

asoong:
One case in RebalancingSetToken#setUpAuctionParameters I ignored for now due to stack size errors. Going to address post refactor when we apply other optimizations (not having to declare contract instances of `currentSet` and `nextSet`.

### DIFF
--- a/contracts/core/modules/IssuanceOrderModule.sol
+++ b/contracts/core/modules/IssuanceOrderModule.sol
@@ -576,9 +576,12 @@ contract IssuanceOrderModule is
         view
         returns (uint256[])
     {
+        // Storing component count to local variable to save on invocation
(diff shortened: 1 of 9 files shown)


contracts/core/modules/IssuanceOrderModule.sol
```diff
@@ -636,9 +636,12 @@ contract IssuanceOrderModule is
         view
         returns (uint256[])
     {
+        // Storing component count to local variable to save on invocation
+        uint256 requriedComponentCount = _order.requiredComponents.length;
```
felix2feng: `required`

asoong:
Minor, and only in cases with multiple iterations executing

PR #325, opened 2018-12-17
Made issueInVaultInternal function that allows a module to issueInVau…

bweick:
…lt on someone else's behalf. Added to CoreModuleInteraction.

### DIFF
--- a/contracts/core/extensions/CoreIssuance.sol
+++ b/contracts/core/extensions/CoreIssuance.sol
@@ -267,6 +267,7 @@ contract CoreIssuance is
 
     /**
      * Exchange components for Set tokens, accepting any owner
+     * Used in issue, issueTo, and issueInVaultInternal
      *
      * @param  _componentOwner  Address to use tokens from
      * @param  _setRecipient    Address to issue Set to
@@ -348,11 +349,12 @@ contract CoreIssuance is
     }
 
     /**
-     * Converts recipient's components into Set Token's held directly in Vault
+     * Converts recipient's components into Set Tokenss held directly in Vault.
+     * Used in issueInVault
      *
      * @param _recipient    Address to issue to
      * @param _set          Address of the Set
-     * @param _quantity     Number of tokens to redeem
+     * @param _quantity     Number of tokens to issue
      */
     function issueInVaultInternal(
         address _recipient,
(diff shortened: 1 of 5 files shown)


contracts/core/extensions/CoreIssuance.sol
```diff
@@ -355,6 +347,35 @@ contract CoreIssuance is
         );
     }
 
+    /**
+     * Converts recipient's components into Set Token's held directly in Vault
```
felix2feng: `Set Tokens`

contracts/core/extensions/CoreIssuance.sol
```diff
@@ -355,6 +347,35 @@ contract CoreIssuance is
...
+     *
+     * @param _recipient    Address to issue to
+     * @param _set          Address of the Set
+     * @param _quantity     Number of tokens to redeem
```
felix2feng: `to issue`

contracts/core/extensions/CoreModuleInteraction.sol
```diff
@@ -130,6 +130,34 @@ contract CoreModuleInteraction is
...
+    function issueInVaultModule(
+        address _recipient,
+        address _set,
+        uint256 _quantity
+    )
+        external
+        nonReentrant
+    {
+        // Require that only modules can call function
+        require(
```
felix2feng: Wonder if it helps to have a modifier called `onlyModule`

contracts/core/interfaces/ICoreIssuance.sol
```diff
@@ -44,6 +44,20 @@ contract ICoreIssuance {
     )
         internal;
 
+    /**
+     * Converts recipient's components into Set Token's held directly in Vault
+     *
+     * @param _recipient    Address to issue to
+     * @param _set          Address of the Set
+     * @param _quantity     Number of tokens to redeem
```
felix2feng: Same re: comments

contracts/core/extensions/CoreModuleInteraction.sol
bweick: We've generally gone the pattern of eschewing modifiers since they tend to bloat the contract size.

PR #331, opened 2018-12-19
Improve compilation and remove redundancy in test setup

felix2feng:
Result: The time to start the first unit tests is improved from 3 min+ to 1 min 20 seconds using a standard Macbook Pro.

Changes
- Improve compile times using Docker binaries ([article](https://hackernoon.com/compile-solidity-smart-contracts-faster-in-truffle-projects-eb8bebb61ec3))
- Remove redundant compile when we are already migrating


felix2feng:
@asoong Does it make sense at this point to set up docker for our CI?

PR #334, opened 2018-12-26
Brian/rebalancing refactor settle rebalance

bweick:


contracts/core/interfaces/ICore.sol
```diff
@@ -133,6 +133,18 @@ interface ICore {
     )
         external;
 
+    /**
+     * Converts user's components into Set Token's held directly in Vault
```
asoong: `Set Tokens`

asoong: What does held directly in Vault mean? Who does the SetToken show is the owner and what does the Vault balances show?

PR #335, opened 2019-01-04
Remove fees

asoong:


contracts/core/lib/RebalancingHelperLibrary.sol
```diff
@@ -235,45 +235,4 @@ library RebalancingHelperLibrary {
...
-    function calculateFeeSplit(
-        uint256 _totalFees,
-        ICore _coreInstance
-    )
-        internal
-        view
-        returns (uint256, uint256)
-    {
-        uint256 protocolFee = _totalFees.mul(_coreInstance.protocolFee())
-            .div(BASIS_POINTS_DIVISOR);
```
felix2feng: Can you remove the constant declared above?

PR #338, opened 2019-01-06
Added withdrawFromFailedRebalance function. Added fxns to IRebalancin…

bweick:
…gSetToken.

### DIFF
--- a/contracts/core/modules/RebalanceAuctionModule.sol
+++ b/contracts/core/modules/RebalanceAuctionModule.sol
@@ -35,6 +35,7 @@ import { IVault } from "../interfaces/IVault.sol";
 contract RebalanceAuctionModule is
     ReentrancyGuard
 {
+    using SafeMath for uint256;
 
     /* ============ State Variables ============ */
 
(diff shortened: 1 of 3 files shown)


contracts/core/modules/RebalanceAuctionModule.sol
```diff
@@ -138,4 +139,61 @@ contract RebalanceAuctionModule is
             _quantity
         );
     }
+
+    /**
+     * If a Rebalancing Set Token Rebalance has failed and been put in Drawdown state, user's
+     * can withdraw their portion of the components collateralizing the Rebalancing Set Token.
+     * This burns the user's portion of the Rebalancing Set Token.
```
asoong: Can you specify: in the case of a `paused` token

contracts/core/modules/RebalanceAuctionModule.sol
```diff
@@ -138,4 +139,61 @@ contract RebalanceAuctionModule is
...
+        // Create Rebalancing Set Token instance
+        IRebalancingSetToken rebalancingSetToken = IRebalancingSetToken(_rebalancingSetToken);
+
+        // Make sure the rebalancingSetToken is tracked by Core
+        require(
+            coreInstance.validSets(_rebalancingSetToken),
+            "RebalanceAuctionModule.bid: Invalid or disabled SetToken address"
+        );
+
+        // Get combinedTokenArray from RebalancingSetToken
```
asoong: Did you want to check the state here before doing this?

bweick: There's a state check that occurs in the burn function...so didn't want to get duplicative

PR #339, opened 2019-01-06
Whitelist contract

asoong:


test/contracts/lib/whiteList.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+
+    async function subject(): Promise<string> {
+      return await whiteList.removeAddress.sendTransactionAsync(
+        subjectAddressToRemove,
+        { from: subjectCaller }
+      );
+    }
+
+    it('removes the address from the addresses list', async () => {
```
felix2feng: For my own sanity, can we do a remove where the address is index 0 or 1?

PR #342, opened 2019-01-11
Alex/add price feed contracts

asoong:


contracts/external/DappHub/auth.sol
```diff
@@ -0,0 +1,68 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.4.25;
```
felix2feng: Wonder if it makes sense to use our authorizable contracts instead for consistency

contracts/external/DappHub/feed-factory.sol
```diff
@@ -0,0 +1,33 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.4.25;
+
+import "./price-feed.sol";
+
+contract FeedFactory {
```
felix2feng: Can we just deploy price feeds individually and authorize them in the medianizer instead of us the factory?

felix2feng: Is the goal of importing these contracts just to replicate their set up for Brian to test or to deploy on our own?

contracts/external/DappHub/thing.sol
```diff
@@ -0,0 +1,30 @@
+// thing.sol - `auth` with handy mixins. your things should be DSThings
+
+// Copyright (C) 2017  DappHub, LLC
```
felix2feng: Lol this file is called thing.sol

contracts/external/DappHub/math.sol
```diff
@@ -0,0 +1,84 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.4.25;
+
+contract DSMath {
+    function add(uint x, uint y) internal pure returns (uint z) {
+        require((z = x + y) >= x);
```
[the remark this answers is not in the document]
asoong: All from existing dapp hub (maker dao) contracts. Adding in now for availability in testing for our manager proposal contract. Deployment strategy may change @felix2feng

contracts/external/DappHub/feed-factory.sol
asoong: What would the difference be if we had to approve them in the medianizer? I want to change as little of the contracts as possible since those were audited and what is currently deployed

PR #345, opened 2019-01-13
Add oracle wrapper and tests for sample usage

asoong:
You should be able to use `updatePriceFeedAsync`. Make sure to use read instead of peek which ensures the price has not expired.


package.json
```diff
@@ -91,7 +91,7 @@
     "lint-staged": "^7.2.0",
     "module-alias": "^2.1.0",
     "openzeppelin-solidity": "^2.0.0",
-    "set-protocol-utils": "^1.0.0-beta.22",
+    "set-protocol-utils": "^1.0.0-beta.23",
```
felix2feng: Let's use 0.24 since its the latest and to ensure no collisions between my PR and yours

test/contracts/external/median.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+ChaiSetup.configure();
+const web3 = getWeb3();
+const FeedFactory = artifacts.require('FeedFactory');
+const { SetProtocolTestUtils: SetTestUtils, SetProtocolUtils: SetUtils } = setProtocolUtils;
+const { expect } = chai;
+const blockchain = new Blockchain(web3);
+const setUtils = new SetUtils(web3);
+
+
+contract('Oracle', accounts => {
```
felix2feng: Make this consistent with the contract name or an extension of it? `Median Oracle`?

test/contracts/external/median.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      priceFeed = await oracleWrapper.deployPriceFeedAsync(feedFactory);
+    });
+
+    describe('#poke', async () => {
+      let subjectPrice: BigNumber;
+      let subjectTimestamp: BigNumber;
+      let subjectCaller: Address;
+
+      beforeEach(async () => {
+        subjectPrice = new BigNumber(10000000);
```
felix2feng: Suggestion: use a realistic price? I believe they use 18 decimal places. So $100 is 100 * 10^18.

test/contracts/external/median.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      let subjectTimestamps: BigNumber[];
+      let subjectVSignatures: BigNumber[];
+      let subjectRSignatures: Bytes[];
+      let subjectSSignatures: Bytes[];
+      let subjectCaller: Address;
+
+      let updatedPrice: BigNumber;
+
+      beforeEach(async () => {
+        oracleCount = 3;
```
felix2feng: How does the medianizer handle even numbers of oracles (e.g. 4)?

utils/wrappers/oracleWrapper.ts
```diff
@@ -0,0 +1,141 @@
+import * as _ from 'lodash';
+import * as setProtocolUtils from 'set-protocol-utils';
```
felix2feng: This wrapper file is legit. 💯

utils/wrappers/oracleWrapper.ts
```diff
@@ -0,0 +1,141 @@
...
+  ): Promise<string> {
+    return await priceFeed.poke.sendTransactionAsync(
+      price,
+      timeStamp,
+      { from },
+    );
+  }
+
+  /*
+    This is disconnected from the v1 system where price feeds are updated first and then
```
felix2feng: Great documentation

test/contracts/external/median.spec.ts
asoong: Auto rejects, you must send an odd number in

package.json
asoong: +1

test/contracts/external/median.spec.ts
asoong: +1

test/contracts/external/median.spec.ts
asoong: +1

PR #346, opened 2019-01-14
Redeem rb set into components

felix2feng:
Changelog
- Fix some of the interfaces for ISetToken and IRebalancingSetToken
- Added parameter and renamed `RedeemAndWithdrawTo`
- Wrote `redeemRBSetIntoBaseComponents` function and module
- Unit tested module


contracts/core/modules/RebalancingTokenIssuanceModule.sol
```diff
@@ -0,0 +1,131 @@
...
+        returns (uint256)
+    {
+        // Get Base Set Details from the rebalancing Set
+        uint256 baseSetNaturalUnit = IRebalancingSetToken(_baseSetAddress).naturalUnit();
+        uint256 baseSetBalance = vaultInstance.getOwnerBalance(
+            _baseSetAddress,
+            address(this)
+        );
+
+        require(
```
felix2feng: @asoong Instead of that funky return baseSet, we could do something like a require where the input issue or rebalance quantity needs to work

asoong: Yeah I like this better, with the tradeoff being issuance quantity granularity?

PR #351, opened 2019-01-16
Brian/rebalancing token manager

bweick:


contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+import { RebalancingHelperLibrary } from "../../core/lib/RebalancingHelperLibrary.sol";
+
+
+/**
+ * @title LinearAuctionPriceCurve
+ * @author Set Protocol
+ *
+ * Contract used to manage a Rebalancing Set Token
+ */
+contract RebalancingTokenManager {
```
asoong: As a supplementary contract, lets give this a more precise name, it should denote the assets BTCETHRebalancingManager?

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /* ============ External ============ */
+
+    /*
+     * When allowed on RebalancingSetToken, anyone can call for a new rebalance proposal
+     *
+     * @param  _rebalancingSetTokenAddress     The address of Rebalancing Set Token to propose new allocation
+     */
+    function proposeNewRebalance(
```
asoong: `propose`? makes sense to me that this contract should conform to the existing interface

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+
+        // Get naturalUnit and units of currentSet
+        uint256 currentSetNaturalUnit = currentSetTokenInterface.naturalUnit();
+        uint256[] memory currentSetUnits = currentSetTokenInterface.getUnits();
+
+        // Calculate wbtc dollar value in currentSet (in cents)
+        uint256 btcDollarAmount = calculateTokenAllocationAmountUSD(
+            _btcPrice,
+            currentSetNaturalUnit,
+            currentSetUnits[0],
```
asoong: This means btc always has to be the first component/unit stored?

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+        components[1] = ethAddress;
+
+        // Create the nextSetToken contract that collateralized the Rebalancing Set Token once rebalance
+        // is finished
+        address nextSetAddress = coreInterface.create(
+            setTokenFactory,
+            components,
+            units,
+            naturalUnit,
+            bytes32("btceth"),
```
asoong: Let's name these something better

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+            bytes32("btceth"),
```
asoong: BTCETH?

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
```diff
@@ -0,0 +1,456 @@
...
+        uint256[] _units
+    )
+        private
+        returns (uint256)
+    {
+        // Calculate btcDollarAmount of one Set Token (in cents) 
+        uint256 btcDollarAmount = calculateTokenAllocationAmountUSD(
+            _btcPrice,
+            _naturalUnit,
+            _units[0],
```
asoong: Let's only reference these once and make sure they're consistent, e.g. passing in something more precise like `_bitcoinUnits` into the method. Fetch that once in `calculateNextSetUnits`

migrations/5_supplementary.js
```diff
@@ -0,0 +1,35 @@
...
+  if (network == "development" || network == "coverage") {
+    console.log("Exiting - Network is development");
+    return;
+  }
+
+  deployer.then(() => deploySupplmentContracts(deployer, network));
+};
+
+async function deploySupplmentContracts(deployer, network) {
+  // Deploy RebalancingTokenManager
```
asoong: I'd take it up a notch, right now only deploy in development and include a medianizer. Also include a switch statement and different constants for each environment. Can be addressed in separate PR

utils/wrappers/rebalancingWrapper.ts
```diff
@@ -495,4 +498,131 @@ export class RebalancingWrapper {
       priceDenominator,
     };
   }
+
+  /* ============ Rebalancing Token Manager ============ */
+
+  public async deployRebalancingTokenManagerAsync(
```
asoong: probably could be more specific here, also changing the name of the contract

utils/wrappers/rebalancingWrapper.ts
```diff
@@ -495,4 +498,131 @@ export class RebalancingWrapper {
...
+  public getExpectedNextSetParameters(
+    btcPrice: BigNumber,
+    ethPrice: BigNumber,
+  ): any {
+    let units: BigNumber[];
+    let naturalUnit: BigNumber;
+    if (btcPrice.greaterThanOrEqualTo(ethPrice)) {
+      const ethUnits = btcPrice.mul(new BigNumber(10 ** 10)).div(ethPrice).round(0, 3);
+      units = [new BigNumber(1), ethUnits];
+      naturalUnit = new BigNumber(10 ** 10);
```
asoong: Feels like `new BigNumber(10 ** 10)` could be a constant. `DEFAULT_SET_TOKEN_NATURAL_UNIT`

utils/wrappers/rebalancingWrapper.ts
```diff
@@ -495,4 +498,131 @@ export class RebalancingWrapper {
...
+
+  private computeTokenValue(
+    units: BigNumber[],
+    naturalUnit: BigNumber,
+    btcPrice: BigNumber,
+    ethPrice: BigNumber,
+  ): BigNumber {
+    const FULL_TOKEN_AMOUNT = new BigNumber(10 ** 18);
+    const WBTC_TOKEN_DECIMALS = new BigNumber(10 ** 8);
+    const WETH_TOKEN_DECIMALS = new BigNumber(10 ** 18);
```
asoong: Same, all constants here. You should maybe add these to set-protocol-utils so they can be reused.

contracts/supplementary/rebalancing-manager/RebalancingTokenManager.sol
bweick: Yeah, we have to get that right on deployment, but after that it's self-reinforcing

contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
```diff
@@ -0,0 +1,456 @@
...
+            // Calculate btc units as (ethPrice/btcPrice)*100. 100 is used to add 
+            // precision. The increase in unit amounts is offset by increasing the
+            // naturalUnit by two orders of magnitude so that issuance cost is still
+            // roughly the same
+            uint256 ethBtcPrice = _ethPrice.mul(PRICE_PRECISION).div(_btcPrice);
+
+            // Create unit array and define natural unit
+            units[0] = ethBtcPrice; 
+            units[1] = PRICE_PRECISION.mul(DECIMAL_DIFF_MULTIPLIER);
+            naturalUnit = uint256(10**12);
```
felix2feng: Is there a way to solve this without 2 different natural units?

PR #353, opened 2019-01-17
Remove state declaration for rebalancing set token

asoong:


contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -60,10 +60,6 @@ contract RebalancingSetToken is
     uint256 constant MAX_AUCTION_TIME_TO_PIVOT = 259200;
 
     /* ============ Enums ============ */
-
-    enum State { Default, Proposal, Rebalance }
-
-    /* ============ State Variables ============ */
```
felix2feng: Remove Enum separator and not State variable?

PR #355, opened 2019-01-18
Added OpenLinearAuctionPriceCurve which allows for the auction start …

bweick:
…price to be set instead of starting at 0.


contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -36,19 +36,22 @@ contract LinearAuctionPriceCurve {
     uint256 constant public THIRTY_SECONDS = 30;
 
     uint256 public priceDenominator;
+    bool public usesStartPrice;
 
     /*
      * Declare price denominator (or precision) of price curve
      *
      * @param  _priceDenominator        The priceDenominator you want this library to always return
```
felix2feng: Add javadoc

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -134,10 +144,13 @@ contract LinearAuctionPriceCurve {
 
         // If time hasn't passed to pivot use the user-defined curve
         if (elapsed <= _auctionParameters.auctionTimeToPivot) {
-            // Calculate the priceNumerator as a linear function of time between 0 and _auctionPivotPrice
-            priceNumerator = elapsed
-                .mul(_auctionParameters.auctionPivotPrice)
-                .div(_auctionParameters.auctionTimeToPivot);
+            // Calculate the priceNumerator as a linear function of time between _auctionStartPrice and
+            // _auctionPivotPrice
+            priceNumerator = auctionStartPrice.add(
```
felix2feng: Is it possible to break up the calculations (particularly the nested ones) for readability sake?

contracts/core/tokens/RebalancingSetTokenFactory.sol
```diff
@@ -167,9 +165,7 @@ contract RebalancingSetTokenFactory {
     /* ============ Private Functions ============ */
 
     function parseRebalanceSetCallData(
-        bytes _callData,
-        bytes32 _name,
-        bytes32 _symbol
+        bytes _callData
```
felix2feng: Thanks for doing this

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -97,6 +100,12 @@ contract LinearAuctionPriceCurve {
         uint256 priceNumerator = _auctionParameters.auctionPivotPrice;
         uint256 currentPriceDenominator = priceDenominator;
 
+        // Determine the auctionStartPrice based on if it should be sef-defined
```
felix2feng: We should add a warning comment about the implications of this for future coders, as this could easily be a source of bugs.

felix2feng: Also `self-defined`

contracts/core/lib/auction-price-libraries/LinearAuctionPriceCurve.sol
```diff
@@ -22,7 +22,7 @@ import { RebalancingHelperLibrary } from "../RebalancingHelperLibrary.sol";
 
 
 /**
- * @title LinearAuctionPriceCurve
+ * @title OpenLinearAuctionPriceCurve
```
felix2feng: Update comment

PR #357, opened 2019-01-18
BTCETHRebalancingManager: Add Proposal Event

felix2feng:
Changelog
- Add LogManagerProposal and associated unit tests
- Constantify some variables

### DIFF
--- a/contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
+++ b/contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
@@ -66,11 +66,8 @@ contract BTCETHRebalancingManager {
     /* ============ Events ============ */
 
     event LogManagerProposal(
-        uint256 auctionStartPrice,
-        uint256 auctionPivotPrice,
         uint256 btcPrice,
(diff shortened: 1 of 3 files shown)


contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
```diff
@@ -158,6 +172,14 @@ contract BTCETHRebalancingManager {
             auctionStartPrice,
             auctionPivotPrice
         );
+
+        emit LogManagerProposal(
+            auctionStartPrice,
+            auctionPivotPrice,
+            btcPrice,
+            ethPrice,
+            currentSetDollarAmount
```
bweick: Do we want to log to currentSetDollarAmount or the nextSetDollarAmount (or both?). I'm not sure the currentSetDollarAmount is as relevant.

felix2feng: Hmmm do we even need to log the dollar amount? It can be implicitly calculated on the backend with the nextSet address and the prices.

PR #369, opened 2019-01-30
Added arrays that track addresses for modules, exchanges, factories, …

bweick:
…and priceLibraries. When removing address revert if not already enabled.

### DIFF
--- a/contracts/core/extensions/CoreInternal.sol
+++ b/contracts/core/extensions/CoreInternal.sol
@@ -174,6 +174,11 @@ contract CoreInternal is
             "CoreInternal.removeExchange: Exchange already removed"
         );
 
+        require(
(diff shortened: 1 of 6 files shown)


contracts/core/extensions/CoreInternal.sol
```diff
@@ -149,18 +160,27 @@ contract CoreInternal is
...
     )
         external
         onlyOwner
     {
-        state.exchanges[_exchangeId] = address(0);
+        require(
+            state.exchangeIds[_exchangeId] != address(0),
+            "CoreInternal.removeExchange: Exchange already removed"
+        );
+
```
felix2feng: Do we need to check that the _exchangeID matches the exchange address? Otherwise can't we put in an arbitrary exchange?

PR #376, opened 2019-02-03
Brian/require issue to start rebalance

bweick:


contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -341,10 +341,21 @@ contract RebalancingSetToken is
     function endFailedAuction()
         external
     {
+        (
+            ,
+            uint256 calculatedUnitShares
+        ) = StandardSettleRebalanceLibrary.calculateNextSetIssueQuantity(
```
felix2feng: Can we write another function in `StandardSettleRebalanceLibrary` called something like `canPeformSuccessfulSettlement` or so instead of passing through calculatedUnitShares?

contracts/core/tokens/rebalancing-libraries/StandardSettleRebalanceLibrary.sol
```diff
@@ -97,6 +97,11 @@ library StandardSettleRebalanceLibrary {
             _vaultAddress
         );
 
+        require(
```
felix2feng: `canPeformSuccessfulSettlement` function result would be more clear

contracts/core/tokens/rebalancing-libraries/StandardStartRebalanceLibrary.sol
```diff
@@ -111,6 +111,13 @@ library StandardStartRebalanceLibrary {
             _vaultAddress
         );
 
+        // Require remainingCurrentSets to be greater than minimumBid otherwise no bidding would
+        // be allowed
+        require(
+            biddingParameters.remainingCurrentSets >= biddingParameters.minimumBid,
```
felix2feng: Can we also have a function in settleLibrary to reflect this equality? It's also in `endFailedAuction`

PR #377, opened 2019-02-05
Add non-operational modifiers to missing functions

felix2feng:


contracts/core/extensions/CoreIssuance.sol
```diff
@@ -283,6 +283,7 @@ contract CoreIssuance is
         uint256 _quantity
     )
         internal
+        whenOperational
```
bweick: Since we're adding this at a lower level I think there's higher level `whenOperational` modifiers we can remove. Also does the same modifier need to be added for `issueInVaultInternal`?

felix2feng: Hmm. true. Actually `issueInVaultInternal` uses issueInternal

contracts/core/extensions/CoreModuleInteraction.sol
```diff
@@ -143,7 +143,6 @@ contract CoreModuleInteraction is
         uint256 _quantity
     )
         external
-        nonReentrant
```
felix2feng: We dont need this as its only callable by a module

PR #379, opened 2019-02-05
Brian/bespoke rebalancing deploy

bweick:


contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
```diff
@@ -58,18 +56,17 @@ contract BTCETHRebalancingManager {
...
 
-    uint256 public btcMultiplier;
-    uint256 public ethMultiplier;
     address public auctionLibrary;
     uint256 public auctionTimeToPivot;
+    uint256 public btcMultiplier;
+    uint256 public ethMultiplier;
+
+    uint256 public lowerAllocationBound;
+    uint256 public upperAllocationBound;
```
felix2feng: Same comment as other PR: Can we call this `minimumUpperThreshold` and `maximumLowerThreshold` to match rebalance strategy terminology?

contracts/supplementary/rebalancing-manager/BTCETHRebalancingManager.sol
```diff
@@ -109,28 +106,32 @@ contract BTCETHRebalancingManager {
         address _setTokenFactory,
         address _auctionLibrary,
         uint256 _auctionTimeToPivot,
-        uint256 _btcMultiplier,
-        uint256 _ethMultiplier
+        uint256[] _multipliers,
```
felix2feng: Would we want to fix this to `uint[2]` and the same w/ allocationBounds?

felix2feng: Random thought. Can't we have a uint256[4] instead of 2 uint256[2]?

Also, consider doing uint64[4] or uint128[2] to pack the numbers together, since we don't expect multipliers or lower/upper bound to be very huge.

PR #383, opened 2019-02-07
RebalancingSetToken: Add naturalUnit param to creation

felix2feng:


contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -156,6 +152,12 @@ contract RebalancingSetToken is
             "RebalancingSetToken.constructor: Unit shares must be positive"
         );
 
+        // Require natural unit is non-zero
+        require(
+            _naturalUnit > 0,
```
bweick: I think this should at least have a lower bound. If not an upper one too. I especially don't like the idea of a rebalancing set with a naturalUnit of 1. It will lead to a lot of failed rebalances I imagine when people try to just use unitShares = 1 on initialization.

test/scenarios/rebalancing-token-issuance/btcEthMultipleRebalance.spec.ts
```diff
@@ -219,7 +220,7 @@ contract('Multiple Rebalance BTC-ETH 50/50', accounts => {
     return btcEthRebalanceWrapper.runFullRebalanceProgram();
   }
 
-  describe('for multiple rebalance cycles', async () => {
+  describe.only('for multiple rebalance cycles', async () => {
```
bweick: remove .only

contracts/core/tokens/RebalancingSetToken.sol
felix2feng: We could throw a minimumNaturalUnit and maximumNatural unit in the factory upon its construction. Maybe something between 10**4 and 10**12

PR #391, opened 2019-02-21
Implement new deployment script

kermankohli:


deployments/README.md
```diff
@@ -0,0 +1,26 @@
+# Getting Started
+
+Before getting started, make sure you have the following `env` variables set:
+
+```
+// Private key to deploy with
+DEPLOYMENT_PRIVATE_KEY="0xA..."
+
+// The network name set here will use constants from `network-constants.ts`
+DEPLOYMENT_NETWORK_NAME="development"
```
felix2feng: Let's detail which networks you can deploy to. `kovan`, `main` etc.

deployments/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+// Network id will indicate which chain to run on (testnet, mainnet etc)
+DEPLOYMENT_NETWORK_ID=50;
```
felix2feng: Which network corresponds to what? Would be helpful

.gitignore
```diff
@@ -29,3 +29,6 @@ yarn-error.log*
 
 # Snapshot
 blockchain/
+
+# Ouputs
+outputs.ts
```
felix2feng: Do you mean deployments/outputs.ts?

kermankohli: Not sure if the path makes a difference but added it for granularity's sake

PR #392, opened 2019-02-22
CoreModuleInteraction Generalization

felix2feng:
- Add module functions: batchIncrementTokenOwnerModule, batchDecrementTokenOwnerModule, batchTransferBalanceModule, transferModule, batchTransferModule

Reducing Bytecode size
- Moved issuance calculations to a library
- Shortened Core error strings to decrease bytecode size

Misc
- Use `onlyModule` modifier in CoreModuleInteraction

### DIFF
--- a/contracts/core/extensions/CoreModuleInteraction.sol
+++ b/contracts/core/extensions/CoreModuleInteraction.sol
@@ -36,6 +36,14 @@ contract CoreModuleInteraction is
     CoreState,
     ReentrancyGuard
 {
+    modifier onlyModule() {
+        require(
+            state.validModules[msg.sender],
+            "Core: Sender not recognized module"
+        );
+        _;
+    }
+
     /**
      * Exposes internal function that deposits multiple tokens to the vault, to system
      * modules. Quantities should be in the order of the addresses of the tokens being
@@ -53,13 +61,8 @@ contract CoreModuleInteraction is
         uint256[] _quantities
     )
         external
+        onlyModule
     {
-        // Require that only modules can call function
-        require(
-            state.validModules[msg.sender],
-            "Core.batchDepositModule: Sender not recognized module"
-        );
-
         batchDepositInternal(
             _from,
             _to,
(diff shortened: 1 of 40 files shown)


deployments/stages/1_libraries.ts
```diff
@@ -68,6 +69,18 @@ export class LibrariesStage implements DeploymentStageInterface {
     return await ExchangeIssueLibraryContract.at(address, this._web3, TX_DEFAULTS);
   }
 
+  private async deployIssuanceLibrary(): Promise<IssuanceLibraryContract> {
```
kermankohli: You'll need to update the tests file to ensure this gets deployed

deployments/stages/4_authorization.ts
```diff
@@ -62,19 +62,8 @@ export class AuthorizationStage implements DeploymentStageInterface {
...
-      'RebalancingTokenIssuanceModule',
-    ]);
-
-    await this.addAuthorizedAddressesToTransferProxy([
-      'Core',
-      'ExchangeIssueModule',
-      'RebalanceAuctionModule',
-      'RebalancingTokenIssuanceModule',
-    ]);
+    await this.addAuthorizedAddressesToVault(['Core']);
```
kermankohli: Tests need to be updated for this too

contracts/core/extensions/CoreModuleInteraction.sol
```diff
@@ -36,6 +36,18 @@ contract CoreModuleInteraction is
     CoreState,
     ReentrancyGuard
 {
+    modifier onlyModule() {
```
bweick: We've kind of eschewed the use of modifiers why start now? Does this help with bytesize? As far as I remember modifiers get compiled into the function anyway so this increases bytesize right?

felix2feng: In this case, it became really un-DRY to have the same modifier code for 8 functions

PR #393, opened 2019-02-23
Upgrade solidity version to 0.5.4

justinkchen:
Running `yarn compile` and gradually fixing changes.

There are a lot of breaking changes with https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html

Following this guide:
https://hackernoon.com/compile-solidity-smart-contracts-faster-in-truffle-projects-eb8bebb61ec3

Example new errors:
https://ethereum.stackexchange.com/questions/8959/solidity-compile-expected-token-comma-got-identifier
https://ethereum.stackexchange.com/questions/63426/running-out-of-gas-during-a-deploy-due-to-a-large-number-of-require-statements
https://github.com/sc-forks/solidity-parser/pull/18
https://hackernoon.com/compile-solidity-smart-contracts-faster-in-truffle-projects-eb8bebb61ec3
`Data location must be "calldata" for parameter in external function, but none was given.`
`Data location must be "storage" or "memory" for parameter in function, but none was given`
`Data location must be "memory" for return parameter in function, but none was given.`


package.json
```diff
@@ -18,10 +18,10 @@
     "test": "test"
   },
   "scripts": {
-    "chain": "yarn clean-chain && ganache-cli --db blockchain --networkId 50 --accounts 20 -l 19000000 -e 10000000000 -m 'concert load couple harbor equip island argue ramp clarify fence smart topic'",
+    "chain": "yarn clean-chain && ganache-cli --db blockchain --networkId 50 --accounts 20 -l 19000000 -e 10000000000 -m 'concert load couple harbor equip island argue ramp clarify fence smart topic' --allowUnlimitedContractSize",
```
felix2feng: Before we merge... let's try to do a test deploy on Kovan and check the gas costs and whether Core could be deployed. This way we can validate whether this flag interferes with deployment

package.json
```diff
@@ -18,10 +18,10 @@
...
     "clean": "rm -rf build; rm -rf transpiled; rm -rf types/generated",
     "clean-chain": "rm -rf blockchain && cp -r snapshots/0x-Kyber blockchain",
-    "compile": "truffle compile --all",
+    "compile": "./node_modules/.bin/truffle compile --all",
```
felix2feng: any reason for this?

package.json
```diff
@@ -79,8 +79,8 @@
     "json-stable-stringify": "^1.0.1",
     "lodash": "^4.17.4",
     "set-abi-gen": "1.1.0-beta.1",
-    "solc": "^0.4.25",
-    "solidity-coverage": "^0.5.11",
+    "solc": "^0.5.4",
+    "solidity-coverage": "https://github.com/leapdao/solidity-coverage#master",
```
felix2feng: Any reason for changing to this?

contracts/core/tokens/SetToken.sol
```diff
@@ -127,7 +127,8 @@ contract SetToken is
 
             // Figure out which of the components has the minimum decimal value
             /* solium-disable-next-line security/no-low-level-calls */
-            if (currentComponent.call(bytes4(keccak256("decimals()")))) {
+            (bool success, ) = currentComponent.call(abi.encodeWithSignature("decimals()"));
```
felix2feng: Interesting new return of .call

contracts/core/tokens/SetTokenFactory.sol
```diff
@@ -84,13 +84,15 @@ contract SetTokenFactory
...
 
         // Create a new SetToken contract
-        return new SetToken(
-            this,
-            _components,
-            _units,
-            _naturalUnit,
-            _name.bytes32ToString(),
-            _symbol.bytes32ToString()
+        return address(
```
felix2feng: why do we need to wrap this in `address`?

contracts/external/DappHub/note.sol
```diff
@@ -34,8 +34,13 @@ contract DSNote {
             bar := calldataload(36)
         }
 
-        emit LogNote(msg.sig, msg.sender, foo, bar, msg.value, msg.data);
```
felix2feng: This is a breaking change? Can't get msg.value directly?

contracts/mocks/tokens/InvalidReturnTokenMock.sol
```diff
@@ -41,13 +41,6 @@ contract InvalidReturnTokenMock {
     decimals = _decimals;
   }
 
-  /**
-  * @dev Total number of tokens in existence
-  */
-  function totalSupply() external view returns (uint256) {
```
felix2feng: any reason this was removed?

package.json
justinkchen: yes i wanted to get this done too, need to make sure we can use the optimizer i think as mentioned here: https://ethereum.stackexchange.com/questions/63426/running-out-of-gas-during-a-deploy-due-to-a-large-number-of-require-statements

package.json
justinkchen: i think this was to use the upgraded version of solc compiler as specified in the `truffle.js`

https://hackernoon.com/compile-solidity-smart-contracts-faster-in-truffle-projects-eb8bebb61ec3

i think before this, it was still trying to use the old version since truffle v5 by default still uses 0.4.25

package.json
justinkchen: https://github.com/sc-forks/solidity-parser/pull/18 coverage was failing because of `_` in variable names and some other weird stuff that the current version of solidity parser doesn't handle still. we can maybe fork leapdaos too if we want to keep it under us for now

contracts/core/tokens/SetTokenFactory.sol
justinkchen: I forget the exact error message, but basically it was saying that we were returning an instance of a contract instead of an address and a lot of the solc 0.5 changes are all about being more explicit so have to convert to address i think

contracts/external/DappHub/note.sol
justinkchen: yea can't use msg.value directly, needs to be in an internal function

justinkchen: `Using msg.value in non-payable functions (or introducing it via a modifier) is disallowed as a security feature.`

contracts/mocks/tokens/InvalidReturnTokenMock.sol
justinkchen: It is already defined as a public variable on line 15 here, so the getter should automatically be generated. it was giving me an error about shadowing variables here

PR #400, opened 2019-02-28
Fix deployment script, add error checking and UX improvements

kermankohli:


deployments/network-constants.ts
```diff
@@ -2,88 +2,88 @@ import constants from './constants';
 
 export default {
   minimumRebalanceInterval: {
-    main: constants.ONE_DAY_IN_SECONDS,
-    kovan: constants.ONE_MINUTE_IN_SECONDS,
+    production: constants.ONE_DAY_IN_SECONDS,
```
asoong: What does production and staging refer to? This could be simplified to just be one setting per environment instead of mapping production/staging to our environments

PR #410, opened 2019-03-04
Add Contract name constants and bump to 0.13

felix2feng:


deployments/contractNames.ts
```diff
@@ -0,0 +1,42 @@
...
+  RebalancingTokenIssuanceModule: 'RebalancingTokenIssuanceModule',
+  SetTokenFactory: 'SetTokenFactory',
+  StandardFailAuctionLibrary: 'StandardFailAuctionLibrary',
+  StandardPlaceBidLibrary: 'StandardPlaceBidLibrary',
+  StandardProposeLibrary: 'StandardProposeLibrary',
+  StandardSettleRebalanceLibrary: 'StandardSettleRebalanceLibrary',
+  StandardStartRebalanceLibrary: 'StandardStartRebalanceLibrary',
+  TransferProxy: 'TransferProxy',
+  Vault: 'Vault',
+  WhiteList: 'WhiteList'  ,
```
asoong: white space

deployments/outputs.json
```diff
@@ -36,5 +36,45 @@
       "last_deployment_stage": 5,
       "network_id": 42
     }
+  },
+  "50-development": {
```
asoong: unstage these in source tree before committing

deployments/outputs.ts
```diff
@@ -37,5 +37,45 @@ export const outputs =
       "last_deployment_stage": 5,
       "network_id": 42
     }
+  },
+  "50-development": {
```
asoong: same

deployments/contractNames.ts
```diff
@@ -0,0 +1,42 @@
+export const CONTRACT = {
+  BitEthInitialCollateralSet: 'BitEthInitialCollateralSet',
```
asoong: This seems really roundabout and error prone. Can we do something like: https://stackoverflow.com/questions/10314338/get-name-of-object-or-class

Import them all from the `./ts` folder and then use this?

deployments/outputs.json
kermankohli: if you run `yarn delete-developments` it should wipe it

PR #411, opened 2019-03-04
Alex/update circle config

asoong:


.env.default
```diff
@@ -1,8 +1,8 @@
 # Private key must be prepended by '0x'
-DEPLOYMENT_PRIVATE_KEY="" 
+DEPLOYMENT_PRIVATE_KEY="0xf2f48ee19680706196e2e339e5da3491186e0c4c5030670656b0e0164837257d"
```
felix2feng: Did we mean to commit this?

asoong: Yes, the default generally has sample values. I'm taking the first pair produced when initializing chain.

PR #413, opened 2019-03-05
Export contract addresses

kermankohli:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "set-protocol-contracts",
-  "version": "1.2.1-beta.13",
+  "version": "1.2.1-beta.14",
   "description": "Smart contracts for {Set} Protocol",
   "main": "dist/artifacts/index.js",
```
felix2feng: Make sure to update `delete-dev-addresses` and also the gitignore might need to be updated?

scripts/generate_outputs.sh
```diff
@@ -8,5 +8,5 @@ rm -f deployments/outputs.ts
 
 filename=outputs
```
felix2feng: Does `rm -f deployments/outputs.ts` need to get updated too?

PR #420, opened 2019-03-08
Generalized BTCETH deployment. Moved deploy token params to contractN…

bweick:
…ames file. Define allocation bounds as difference from ideal allocation, and create helper function to calculate bounds submitted to contract.


deployments/contractNames.ts
```diff
@@ -1,11 +1,183 @@
-export const DEPLOYED_TOKEN = {
-  BitEthInitialCollateralSet: 'BitEthInitialCollateralSet',
-  BitEthRebalancingSetToken: 'BitEthRebalancingSetToken',
-  ETHDaiInitialCollateralSet: 'ETHDaiInitialCollateralSet',
-  ETHDaiRebalancingSetToken: 'ETHDaiRebalancingSetToken',
-  BTCDaiInitialCollateralSet: 'BTCDaiInitialCollateralSet',
-  BTCDaiRebalancingSetToken: 'BTCDaiRebalancingSetToken',
-};
+import BigNumber from 'bignumber.js';
```
felix2feng: Is there a name for this file that reflects what it is better what it is?

types/deployment_stage_interface.ts
```diff
@@ -1,3 +1,26 @@
...
+}
+
+interface NetworkConstants {
+	production: any;
+	staging: any;
+	development: any;
+}
+
+export interface DeployedSetInfo {
+	PRICE_PRECISION: BigNumber;
```
felix2feng: Extra Tab

types/deployment_stage_interface.ts
```diff
@@ -1,3 +1,26 @@
...
   deploy(web3: any): Promise<any>;
+}
+
+interface NetworkConstants {
+	production: any;
+	staging: any;
+	development: any;
+}
+
+export interface DeployedSetInfo {
```
felix2feng: 💯 Good stuff adding the typing

types/deployment_stage_interface.ts
```diff
@@ -1,3 +1,26 @@
+import BigNumber from 'bignumber.js';
+
 export interface DeploymentStageInterface {
   deploy(web3: any): Promise<any>;
+}
+
+interface NetworkConstants {
+	production: any;
```
felix2feng: Think these should be numbers vs any type

bweick: Disagree cause there might be other kinds of constants we want to tweak in the future that aren't limited to numbers. Could be addresses and what not.

PR #426, opened 2019-03-19
PostAudit Renaming and Clean up

felix2feng:
- Rename CoreInternal -> CoreAdmin
- Consolidate array length checks into ArrayValidations.sol
- Remove `removeAuthorizedAddressAtIndex`
- Consolidate isMultipleOfSetNaturalUnit, getSetDetails into SetTokenLibrary
- Reduce Core instantiation in ExchangeIssuanceLibrary
- Create `StandardStartRebalanceLibrary.calculateCombinedUnits` to reduce duplication and stack size in `calculateCombinedUnitArrays`
- Add unit tests for all new functions
- Updated a bunch of comments


contracts/core/tokens/SetToken.sol
```diff
@@ -90,22 +89,10 @@ contract SetToken is
         );
 
         // Confirm an empty _components array is not passed
-        require(
-            componentCount > 0,
-            "SetToken.constructor: Empty components array"
-        );
-
-        // Confirm an empty _quantities array is not passed
-        require(
```
bweick: Where'd the validation for `_quantities` go?

felix2feng: The `validateEqualLength` implicitly validates the unit - since the component length needs to be  > 0 and they both need to be the same

contracts/core/extensions/CoreAdmin.sol
```diff
@@ -147,7 +147,7 @@ contract CoreInternal is
     {
         require(
             state.exchangeIds[_exchangeId] == address(0),
-            "Core: Invalid Id"
+            "Invalid Id"
```
bweick: This feels hacky...I know we're trying to reduce bytecode size and this is a good way to do it but if that's the case we should change our error message return system wide. Cause some of these error messages are duplicates which almost defeats the purpose.

asoong: +1. What's the thinking here? What is the difference in size from doing this? There are different ways to achieve this.

I will say that the current system doesn't provide all that much clarity either with duplicate `Core: Invalid Factory`. Some projects i've worked on have gone with a sequential bug identification i.e. `Core-1`, `Core-2`, etc. though with Core being so large, adding or removing a `require` could result in changes across multiple files. That's why we tried originally to do something like `Filename-MethodName-Error`

asoong:
In the future, let's break up PRs, particularly if they involve naming changes and source changes. This is difficult to digest.

PR #427, opened 2019-03-19
Add dependency addresses for mainnet. Make default deploy gas price 4…

bweick:
… gwei.


deployments/test/4_authorization.spec.ts
```diff
@@ -206,7 +206,7 @@ describe('Deployment: Authorization', () => {
 
     it('core contains kyber network exchange wrapper', async () => {
       const kyberNetworkAddress = await getContractAddress(KyberNetworkWrapper.contractName);
-      expect(kyberNetworkAddress).toContain(kyberNetworkAddress);
+      expect(exchanges).toContain(kyberNetworkAddress);
```
asoong: Side note, do we plan to even deploy the 0x exchange wrapper with our upcoming deploy? I thought we were punting

deployments/utils/blockchain.ts
```diff
@@ -23,7 +23,7 @@ export function getInfuraKey(): string {
 
 export let TX_DEFAULTS = {
   gas: 6700000, // 6.7M
-  gasPrice: 10000000000, // 10 gWei
+  gasPrice: 4000000000, // 4 gWei
```
asoong: Instead of changing this back and forth, it should be passed in with the deploy script

PR #429, opened 2019-03-20
Modify PayableExchangeIssuance into a Module

felix2feng:
TODO: Update deployment scripts


asoong:
Does this even need to be its own module then? Why not add payable methods to ExchangeIssuanceModule?

PR #431, opened 2019-03-20
Add internal function to rebalancingSetToken that clears state after …

bweick:
…rebalance is completed (failure or success).

### DIFF
--- a/contracts/core/tokens/RebalancingSetToken.sol
+++ b/contracts/core/tokens/RebalancingSetToken.sol
@@ -83,6 +83,9 @@ contract RebalancingSetToken is
     RebalancingHelperLibrary.AuctionPriceParameters public auctionParameters;
     RebalancingHelperLibrary.BiddingParameters public biddingParameters;
 
+    // To be used if token put into Drawdown State
+    address[] public failedAuctionWithdrawComponents;
+
     /* ============ Events ============ */
 
     event NewManagerAdded(
@@ -368,6 +371,11 @@ contract RebalancingSetToken is
 
         // Reset lastRebalanceTimestamp to now
         lastRebalanceTimestamp = block.timestamp;
+
+        // Save combined token arrays to failedAuctionWithdrawComponents
+        failedAuctionWithdrawComponents = biddingParameters.combinedTokenArray;
+
+        // Clear auction state
         clearAuctionState();
     }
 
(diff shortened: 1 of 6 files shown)


contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -619,4 +621,24 @@ contract RebalancingSetToken is
...
+        startingCurrentSetAmount = 0;
+        delete auctionParameters;
+
+        if (rebalanceState == RebalancingHelperLibrary.State.Drawdown) {
+            biddingParameters.minimumBid = 0;
+            biddingParameters.remainingCurrentSets = 0;
+            delete biddingParameters.combinedCurrentUnits;
+            delete biddingParameters.combinedNextSetUnits;
+        } else {
+            delete biddingParameters;
```
felix2feng: What's the difference between the two cases? Don't they do the same thing?

contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -619,4 +621,24 @@ contract RebalancingSetToken is
     {
         return biddingParameters.combinedNextSetUnits;
     }
+
+    /* ============ Getter Functions ============ */
+
+    function clearAuctionState()
```
felix2feng: Add java doc

contracts/core/tokens/RebalancingSetToken.sol
```diff
@@ -619,4 +621,24 @@ contract RebalancingSetToken is
     {
         return biddingParameters.combinedNextSetUnits;
     }
+
+    /* ============ Getter Functions ============ */
```
felix2feng: Fix naming

contracts/core/tokens/RebalancingSetToken.sol
bweick: We need the `combinedTokenArray` for the `withdrawFromFailedRebalance`

felix2feng: Hmm - why don't we delete `biddingParameters` and generate a variable called `withdrawComponents` or something for withdrawing?

felix2feng: @asoong What do you think?

asoong: Yeah not a fan of this, also not a fan of duplicating the state. Ideally we store them separately and only delete `minimumBid` and `remainingCurrentSets` but i'm guessing they're grouped together now so they can be passed around more efficiently

asoong: Actually, they're only passed together when placing bid and getting bid price

PR #433, opened 2019-03-21
Return non-exchanged components in ExchangeIssuance

felix2feng:


contracts/core/modules/ExchangeIssuanceModule.sol
```diff
@@ -277,4 +280,32 @@ contract ExchangeIssuanceModule is
...
+        address _setAddress
+    )
+        private
+    {
+        address[] memory baseSetComponents = ISetToken(_setAddress).getComponents();
+        uint256[] memory baseSetWithdrawQuantities = new uint256[](baseSetComponents.length);
+        for (uint256 i = 0; i < baseSetComponents.length; i++) {
+            uint256 withdrawQuantity = vaultInstance.getOwnerBalance(baseSetComponents[i], address(this));
+            
+            baseSetWithdrawQuantities[i] = withdrawQuantity;
```
asoong: Check for zeros here?

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -367,4 +370,26 @@ contract RebalancingSetExchangeIssuanceModule is
...
+     *
+     * @param  _setAddress   Address of the Base Set
+     */
+    function returnExcessBaseSetComponents(
+        address _setAddress
+    )
+        private
+    {
+        address[] memory baseSetComponents = ISetToken(_setAddress).getComponents();
+        for (uint256 i = 0; i < baseSetComponents.length; i++) {
```
bweick: Why do we do individual transfers here when we batch them in the `ExchangeIssuanceModule` contract?

bweick: Nvm we don't have a batch transfer functionality

PR #435, opened 2019-03-21
Rebalancing Proposal Cleanup

felix2feng:
- Decouple validation and generation of priceParams
- Rename `auctionParameters` to `auctionPriceParameters`

### DIFF
--- a/contracts/core/tokens/rebalancing-libraries/ProposeLibrary.sol
+++ b/contracts/core/tokens/rebalancing-libraries/ProposeLibrary.sol
@@ -38,10 +38,12 @@ library ProposeLibrary {
 
     /* ============ Structs ============ */
 
-    struct ProposeAuctionParameters {
+    struct ProposalContext {
         address manager;
         address currentSet;
         address coreAddress;
+        address componentWhitelist;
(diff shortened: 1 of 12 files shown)


contracts/core/tokens/rebalancing-libraries/ProposeLibrary.sol
```diff
@@ -50,110 +52,91 @@ library ProposeLibrary {
...
+    function validateProposal(
         address _nextSet,
         address _auctionLibrary,
-        uint256 _auctionTimeToPivot,
-        uint256 _auctionStartPrice,
-        uint256 _auctionPivotPrice,
-        address _factoryAddress,
-        address _componentWhiteListAddress,
-        ProposeAuctionParameters memory _proposeParameters
+        ProposalContext memory _proposalContext,
```
asoong: `Context`..? Prefer `AuctionParameters`

felix2feng: Well.. the struct has the data that the function needs to be aware of and isn't auction-specific.

PR #438, opened 2019-03-21
Rename denominator to divisor for consistency

felix2feng:


contracts/mocks/core/lib/ConstantAuctionPriceCurve.sol
```diff
@@ -35,23 +35,23 @@ contract ConstantAuctionPriceCurve {
...
     uint256 public priceNumerator;
-    uint256 public priceDenominator;
+    uint256 public priceDivisor;
 
     /*
      * Declare price you want this library to return when queried
      *
      * @param  _priceNumerator          The priceNumerator you want this library to always return
-     * @param  _priceDenominator        The priceDenominator you want this library to always return
+     * @param  _priceDivisor        The priceDivisor you want this library to always return
```
asoong: Line these next time

utils/constants.ts
```diff
@@ -3,7 +3,7 @@ import { ether } from '../utils/units';
 
 export const AUCTION_TIME_INCREMENT = new BigNumber(30); // Unix seconds
 export const DEFAULT_AUCTION_PRICE_NUMERATOR = new BigNumber(1374);
-export const DEFAULT_AUCTION_PRICE_DENOMINATOR = new BigNumber(1000);
+export const DEFAULT_AUCTION_PRICE_DIVISOR = new BigNumber(1000);
```
asoong: These should eventually be funned down to set-protocol-utils. Not sure when we stopped doing this to iterate faster on contracts

PR #440, opened 2019-03-21
Optimize vault functions

felix2feng:


bweick:
I'd remove the `quantity > 0` check at the batch level if that's going to exist at the individual level.

PR #454, opened 2019-03-26
Removed unused contract from artifacts index. Updated btc and eth pri…

bweick:
…ces. Removed deploy testing from deploy script, must be called separately now.

### DIFF
--- a/deployments/constants.ts
+++ b/deployments/constants.ts
@@ -7,14 +7,14 @@ export default {
   },
 
   WBTC: {
-    PRICE: new BigNumber(3711),
+    PRICE: new BigNumber(3964),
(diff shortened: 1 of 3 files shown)


deployments/constants.ts
```diff
@@ -7,14 +7,14 @@ export default {
   },
 
   WBTC: {
-    PRICE: new BigNumber(3711),
```
felix2feng: Can you also check other areas where this is the constant (e.g. tests)?

asoong:
Do we still want this PR? For the constant changes?

bweick:
Yeah and the small change to the package.json. I say we just do it.

PR #465, opened 2019-04-04
TimeLockUpgrade: Allow setting of timelock period as long as > than existing

felix2feng:


contracts/lib/TimeLockUpgrade.sol
```diff
@@ -104,8 +104,8 @@ contract TimeLockUpgrade is
         external
         onlyOwner
     {
-        // Only allow setting of the timeLockPeriod the first time
-        if (timeLockPeriod == 0) {
+        // Only allow setting of the timeLockPeriod if the period is greater than the existing
+        if (_timeLockPeriod > timeLockPeriod) {
```
asoong: This is something we should be reverting on instead of a silent failure

felix2feng: +1

PR #481, opened 2019-05-12
Bump web3 version in set-protocol-contracts

justinkchen:


justinkchen:
failing due to this ganache bug: https://github.com/trufflesuite/ganache-core/issues/339

PR #486, opened 2019-06-12
ERC20 Rebalancing Set Exchange Issuance

felix2feng:


test/contracts/core/modules/erc20RebalancingSetExchangeIssuanceModule.spec.ts
```diff
@@ -0,0 +1,912 @@
...
+      };
+
+      await erc20Wrapper.approveTransfersAsync(
+        [baseSetComponent],
+        SetTestUtils.ZERO_EX_ERC20_PROXY_ADDRESS,
+        zeroExOrderMaker
+      );
+
+      // Create 0x order for the component, using weth(4) paymentToken as default
+      zeroExOrder = await setUtils.generateZeroExSignedFillOrder(
```
asoong: same as before, this test is not very indicative of what we're going to launch with.
1. the set has a single component
2. it's a single 0x order
3. it's 0x
at the very minimum lets get these switched over to using kyber for the one component

asoong:
Let's make sure to keep comments updated. Very misleading for auditors

test/contracts/core/modules/erc20RebalancingSetExchangeIssuanceModule.spec.ts
asoong: we have a second to breath now, we should circle back and address this the right way this time.

contracts/core/modules/ERC20RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -0,0 +1,301 @@
...
+    RebalancingSetExchangeIssuance,
+    ReentrancyGuard
+{
+    using SafeMath for uint256;
+
+    /* ============ State Variables ============ */
+
+    // Address and instance of Transfer Proxy contract
+    address public transferProxy;
+    ITransferProxy public transferProxyInstance;
```
bweick: It appears from the ABDK review that these interfaces can also function as an address. So when we store the address and the instance we're just duping the data so maybe it makes sense to remove `transferProxy` and `exchangeIssuanceModule`.

contracts/core/modules/ERC20RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /**
+     * Redeems a Rebalancing Set into a specific ERC20 / receive token. The Rebalancing Set 
+     * is redeemed into the Base Set, and Base Set components are traded for the ERC20 which is sent to the caller.
+     *
+     * @param  _rebalancingSetAddress    Address of the rebalancing Set
+     * @param  _rebalancingSetQuantity   Quantity of rebalancing Set to redeem
+     * @param  _exchangeIssuanceParams   Struct containing data around the base Set issuance
+     * @param  _orderData                Bytecode formatted data with exchange data for acquiring base set components
```
bweick: selling base set components

contracts/core/modules/ERC20RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -0,0 +1,301 @@
...
+        );
+    }
+
+    /**
+     * Redeems a Rebalancing Set into a specific ERC20 / receive token. The Rebalancing Set 
+     * is redeemed into the Base Set, and Base Set components are traded for the ERC20 which is sent to the caller.
+     *
+     * @param  _rebalancingSetAddress    Address of the rebalancing Set
+     * @param  _rebalancingSetQuantity   Quantity of rebalancing Set to redeem
+     * @param  _exchangeIssuanceParams   Struct containing data around the base Set issuance
```
bweick: redemption

contracts/core/modules/ERC20RebalancingSetExchangeIssuanceModule.sol
felix2feng: +1 that is correct

PR #487, opened 2019-06-24
Rebalancing Set Issuance Module

felix2feng:
RebalancingSetIssuanceModule: A smart contract that allows the issuing and redeeming of a Set using ether and/or the user’s existing tokens. With each function, have a boolean option to keep change or send to the Vault. 
Functions:
- issueRebalancingSet: Issue rebalancing Set with the user’s existing tokens in their wallet. Tokens must be approved to the TransferProxy.
- issueRebalancingSetWithEther: Sends ether to the contract, wraps it, and then issues the Rebalancing Set.
- redeemRebalancingSet: Redeems a Rebalancing Set into its base comments
- redeemRebalancingSetWithEther: Any Set that has wrapped Ether in it will be unwrapped before it is sent to the user.

Things to Know:
- During issuance, `getBaseSetIssuanceRequiredQuantity` is called to calculate the implied Base Set quantity. If the quantity is not a multiple of the natural unit, the quantity is rounded up.
- The entire module is excluded from solcover, because the `msg.sender.transfer` which breaks coverage for the entire file. That said, all the flows are tested.
- Refer to the `RebalancingSetIssuanceModule Token Flow Diagrams` in the google docs to get a sense of where the assets are moving.

[ x ] Tested flows with multiple components, including wEth
[ x ] Read through files and scanned comments


contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,455 @@
...
+        );
+
+        // Issue rebalancing Set to the sender
+        coreInstance.issueTo(
+            msg.sender,
+            _rebalancingSetAddress,
+            _rebalancingSetQuantity
+        );
+
+        // Return any excess base Set token to the sender
```
bweick: maybe add comment about either in vault or in wallet depending on boolean indicator

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,455 @@
...
+                msg.sender,
+                address(this),
+                currentComponent,
+                currentComponentQuantity
+            );
+        }
+    }
+
+    /**
+     * During redemption, withdraw the required quantity of Base Set, unwrapwrap Ether, and withdraw
```
bweick: unwrapwrap -> unwrap

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,455 @@
...
+       // Loop through the base Set components and deposit components 
+        for (uint256 i = 0; i < baseSetComponents.length; i++) {
+            address currentComponent = baseSetComponents[i];
+            uint256 currentUnit = baseSetUnits[i];
+
+            // Calculate required component quantity
+            uint256 currentComponentQuantity = _baseSetQuantity.mul(currentUnit).div(baseSetNaturalUnit);
+
+            // If address is weth, deposit weth and transfer eth
+            if (currentComponent == address(weth)) {
```
asoong: Prefer that there is a requirement that one of the required components is WETH and that ETH is wrapped and used. Maybe have this check in the very beginning. When proxy behaviors exist such as being able to use this as basically the same thing as `issueRebalancingSetWithEther` when it was not intended are when attack vectors surface.

contracts/core/modules/lib/RebalancingSetIssuance.sol
```diff
@@ -0,0 +1,203 @@
...
+
+
+/**
+ * @title RebalancingSetIssuance
+ * @author Set Protocol
+ *
+ * The RebalancingSetIssuance contains utility functions used in RebalancingSet 
+ * issuance
+ */
+contract RebalancingSetIssuance is
```
asoong: Where are tests for these? Prefer they be tested separately instead of by proxy through the module. Will allow you to reuse them with confidence.

Also, you won't leave the person who changes the `RebalancingSetIssuanceModule` contract holding the bag when the module contract changes and one of these is no longer used.

test/contracts/core/modules/rebalancingSetIssuanceModule.spec.ts
```diff
@@ -0,0 +1,874 @@
...
+    let customRebalancingUnitShares: BigNumber;
+    let customRebalancingSetQuantity: BigNumber;
+
+    beforeEach(async () => {
+      subjectCaller = functionCaller;
+
+      baseSetComponent = await erc20Wrapper.deployTokenAsync(subjectCaller);
+      await erc20Wrapper.approveTransferAsync(baseSetComponent, transferProxy.address, subjectCaller);
+
+      // Create the Set (1 component)
```
asoong: Let's do two components since the logic between `issueRebalancingSet` and `issueRebalancingSetWithEther` don't flow through each other. Want to make sure all the for loops in the transaction are working properly.

test/contracts/core/modules/rebalancingSetIssuanceModule.spec.ts
```diff
@@ -0,0 +1,874 @@
...
+        subjectRebalancingSetAddress,
+        subjectCaller,
+        subjectRebalancingSetQuantity,
+        rebalancingTokenIssuanceModule.address
+      );
+
+      await SetTestUtils.assertLogEquivalence(formattedLogs, expectedLogs);
+    });
+
+    describe('when the eth sent is more than required', async () => {
```
asoong: Test case for when less is sent than required?

test/contracts/core/modules/rebalancingSetIssuanceModule.spec.ts
```diff
@@ -0,0 +1,874 @@
...
+    let customRebalancingSetIssueQuantity: BigNumber;
+    let customBaseComponentUnit: BigNumber;
+
+    beforeEach(async () => {
+      subjectCaller = functionCaller;
+
+      baseSetComponent = await erc20Wrapper.deployTokenAsync(ownerAccount);
+      await erc20Wrapper.approveTransferAsync(baseSetComponent, transferProxy.address);
+
+      // Create the Set (1 component)
```
asoong: Same, two components for testing the for loops in the flow.

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,455 @@
...
+     * into wrapped Ether and utilized in issuance.
+     * The Base Set is then issued and reissued into the Rebalancing Set. Read more about base Set quantity
+     * in the issueRebalancingSet function.
+     *
+     * @param  _rebalancingSetAddress    Address of the rebalancing Set to issue
+     * @param  _rebalancingSetQuantity   The issuance quantity of Rebalancing Set
+     * @param  _keepChangeInVault        Boolean signifying whether excess base Set is transfered to the user 
+     *                                     or left in the vault
+     */
+    function issueRebalancingSetWithEther(
```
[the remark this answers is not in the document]
felix2feng: I'm okay with `payableIssueRebalancingSet`, but what would the redeem version be? We can't prefix with `payable` since it's not a payable function.

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,455 @@
...
+            _rebalancingSetAddress,
+            _rebalancingSetQuantity
+        );
+
+        // Calculate the Base Set Redeem quantity
+        address baseSetAddress = IRebalancingSetToken(_rebalancingSetAddress).currentSet();
+        uint256 baseSetRedeemQuantity = getBaseSetRedeemQuantity(baseSetAddress);
+
+        // Withdraw base Set to this contract
+        coreInstance.withdraw(
```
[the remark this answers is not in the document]
felix2feng: We don't have a module function that redeems the user's Sets and withdraws to the recipient's address. `redeemModule` is all we exposed.

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,475 @@
...
+    /**
+     * Issue a Rebalancing Set using the base components of the base SetToken.
+     * The base SetToken is then issued into the Rebalancing Set. The Base SetToken quantity issued is calculated
+     * by taking the Rebalancing Set's quantity, unit shares, and natural unit. If the calculated quantity is not
+     * a multiple of the natural unit of the base Set, the quantity is rounded up to the base Set natural unit.
+     *
+     * @param  _rebalancingSetAddress    Address of the rebalancing Set to issue
+     * @param  _rebalancingSetQuantity   The issuance quantity of Rebalancing Set
+     * @param  _keepChangeInVault        Boolean signifying whether excess base Set is transfered to the user 
+     *                                     or left in the vault
```
asoong: What's going on here with this?

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,475 @@
...
+        external
+        payable
+        nonReentrant
+    {
+        // Validate the RebalancingSet is valid and the quantity is a multiple of the natural unit
+        validateRebalancingSetIssuance(_rebalancingSetAddress, _rebalancingSetQuantity);
+
+        address baseSetAddress = IRebalancingSetToken(_rebalancingSetAddress).currentSet();
+
+        validateWethIsAComponentOfSet(baseSetAddress, address(weth));
```
asoong: nit: WETH

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -0,0 +1,475 @@
...
+            _rebalancingSetAddress,
+            msg.sender,
+            _rebalancingSetQuantity
+        );
+    }
+
+    /**
+     * Issue a Rebalancing Set using the base components and ether of the Base Set. The ether is wrapped 
+     * into wrapped Ether and utilized in issuance.
+     * The Base Set is then issued and reissued into the Rebalancing Set. Read more about base Set quantity
```
asoong: Can we scrub all of these `Base Set` usages? `base SetToken`?

contracts/core/modules/RebalancingSetIssuanceModule.sol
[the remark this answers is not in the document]
asoong: I dont think they need to be uniform. Maybe `issueRebalancingSetWrappingEther` `redeemRebalancingSetUnwrappingEther`

contracts/core/modules/RebalancingSetIssuanceModule.sol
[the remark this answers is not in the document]
asoong: +1

contracts/core/modules/RebalancingSetIssuanceModule.sol
felix2feng: 2 line comment?

PR #490, opened 2019-07-05
Consolidated and clarified some logic. Added more thorough javadocs.

bweick:


contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -409,17 +454,15 @@ contract RebalancingSetIssuanceModule is
                     transferProxy,
                     currentComponentQuantity
                 );
-
```
felix2feng: Make a note that Core will automatically deposit WETH during issue

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -364,6 +343,72 @@ contract RebalancingSetIssuanceModule is
 
     /* ============ Private Functions ============ */
 
+    /**
+     * Gets base set address from rebalancing set token and calculates amount of base set needed
+     * for issuance.
+     *
+     * @param  _rebalancingSetAddress    Address of the rebalancing SetToken to issue
+     * @param  _rebalancingSetQuantity   The Quantity of the rebalancing SetToken to issue
+     * @return baseSetAddress            The address of rebalancing set's base set
```
felix2feng: `base SetToken`

contracts/core/modules/RebalancingSetIssuanceModule.sol
```diff
@@ -364,9 +343,75 @@ contract RebalancingSetIssuanceModule is
...
+        address _rebalancingSetAddress,
+        address _baseSetAddress,
+        uint256 _rebalancingSetQuantity,
+        bool _keepChangeInVault
+    )
+        internal
+    {
+        // Issue rebalancing SetToken to the sender
+        coreInstance.issueTo(
+            msg.sender,
```
3sGgpQ8H: This internal function depends on msg.sender as implicit parameter, which makes code harder to read.  Consider making tthis parameter explicit.

PR #492, opened 2019-07-18
Redo Rebalancing Set Exchange Issuance Module

felix2feng:
Tip: You're going to watch to review this PR using split view.


contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -134,313 +139,396 @@ contract RebalancingSetExchangeIssuanceModule is
...
+            baseSetAddress,
+            transferProxy,
+            _keepChangeInVault
+        );
+
+        // Return any extra components acquired during exchangeIssue to the user
+        returnExcessComponentsFromVault(baseSetAddress);
+
+        // Note: paymentTokenQuantity could be spoofed
+        emit LogPayableExchangeIssue(
```
felix2feng: Do we want to log the quantity if it can be spoofed?

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -134,313 +139,396 @@ contract RebalancingSetExchangeIssuanceModule is
...
+        // If keepChangeInVault is true, the baseSetToken is held in the Vault
+        // which is a UX improvement
+        returnExcessBaseSetFromContract(
+            baseSetAddress,
+            transferProxy,
+            _keepChangeInVault
+        );
+
+        // Return any extra components acquired during exchangeIssue to the user
+        returnExcessComponentsFromVault(baseSetAddress);
```
bweick: Do we want to use the _keepChangeInVault flag here?

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
bweick: Maybe we log the `_rebalancingSetQuantity`? Feels more relevant and like you said can't be manipulated.

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
```diff
@@ -134,313 +137,452 @@ contract RebalancingSetExchangeIssuanceModule is
...
+     * Issue a Rebalancing Set using a specified ERC20 payment token. The payment token is used in ExchangeIssue
+     * to acquire the base SetToken components and issue the base SetToken. The base SetToken is then used to
+     * issue the Rebalancing SetToken. The payment token can be utilized as a component of the base SetToken.
+     * All remaining tokens / change are flushed and returned to the user.
      *
-     * @param  _rebalancingSetAddress    Address of the rebalancing Set
-     * @param  _rebalancingSetQuantity   Quantity of rebalancing Set to redeem
-     * @param  _collateralSetAddress     Address of base Set in ExchangeIssueanceParams
-     * @param  _collateralSetAddress     Address of base Set in ExchangeIssueanceParams
+     * Note: We do not validate the rebalancing SetToken quantity and the exchangeIssuanceParams base SetToken
```
bweick: What do you mean by this note?

package.json
```diff
@@ -41,7 +41,7 @@
     "test-nocompile": "yarn transpile && yarn truffle-test-contracts",
     "test-continuous": "yarn deploy-development && truffle test",
     "transpile": "tsc",
-    "truffle-test-contracts": "truffle test `find transpiled/test/contracts -name '*.spec.js'`",
+    "truffle-test-contracts": "truffle test `find transpiled/test/contracts -name 'rebalancingSetExchangeIssuanceModule.spec.js'`",
```
bweick: Remove

utils/contract_logs/rebalancingSetExchangeIssuanceModule.ts
```diff
@@ -4,7 +4,8 @@ import { BigNumber } from 'bignumber.js';
 export function LogPayableExchangeIssue(
   rebalancingSetAddress: Address,
   callerAddress: Address,
-  etherQuantity: BigNumber,
+  paymentToken: Address,
+  paymentTokenQuantity: BigNumber,
```
bweick: Reminder if we switch to logging the `rebalancingSetQuantity` to change here

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
felix2feng: It means that we don't check that the implied base SetToken quantity from the RB Set Quantity == the passed in base SetToken quantity from exchangeIssuanceParams

contracts/core/modules/RebalancingSetExchangeIssuanceModule.sol
felix2feng: Right now, I lean towards not adding the flag - but I'm a bit conflicted. The original problem we wanted to fix was the base SetToken vs. components. If it does become a huge problem, we can add it later and redeploy

PR #495, opened 2019-07-26
Kyber Customized Reserve Updates

felix2feng:
Final Touches:
- Add comments to kyberReserveWrapper

Next Steps:
- Clean up SetUtils and update any upstream setProtocol.js tests as needed.


utils/wrappers/kyberNetworkWrapper.ts
```diff
@@ -0,0 +1,297 @@
...
+import { ConversionRateABI } from '../external/abis/ConversionRateABI';
+import { KyberNetworkABI } from '../external/abis/KyberNetworkABI';
+import { KyberReserveABI } from '../external/abis/KyberReserveABI';
+
+import { KYBER_CONTRACTS, KYBER_PERMISSIONED_ACCOUNTS } from '../kyberSnapshotAddresses';
+
+const web3 = getWeb3();
+
+
+export class KyberNetworkWrapper {
```
bweick: Nit but can we call this and the file as a whole `KyberNetworkConfigWrapper`? Or something similar. Since we have a contract of the same name it's just easier to differentiate.

asoong: +1 I think we should start just calling these `Helper` classes. The wrapper suffix is confusing in general. Was clear when I was working in the repo a lot but now in review it causes slight context hiccups

felix2feng: Ok - I can create a trello card to do the overall refactor.

PR #499, opened 2019-08-14
Add Kyber Test to RebalancingSetExchangeIssuanceModule

felix2feng:
The testing uncovered the fact that during issuance, we might want to allow multiple sendTokens, as each one is needed to send to a different exchangeWrapper. This is why we relaxed the condition that there must be a single transact token.

For exchange redeem, we implicitly validate that you can only have 1 receive token as exchangeIssuance checks that receiveTokens do not have duplicates.

- This PR does significant cleanup and variable naming of the rebalancingSetExchangeIssue tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "test-nocompile": "yarn transpile && yarn truffle-test-contracts",
     "test-continuous": "yarn deploy-development && truffle test",
     "transpile": "tsc",
-    "truffle-test-contracts": "truffle test `find transpiled/test/contracts -name '*.spec.js'`",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "test-nocompile": "yarn transpile && yarn truffle-test-contracts",
     "test-continuous": "yarn deploy-development && truffle test",
     "transpile": "tsc",
-    "truffle-test-contracts": "truffle test `find transpiled/test/contracts -name '*.spec.js'`",
+    "truffle-test-contracts": "truffle test `find transpiled/test/contracts -name 'rebalancingSetExchangeIssuanceModule.spec.js'`",
```
bweick: Remove

PR #500, opened 2019-08-17
Rebalancing exchange issuance post audit improvements

felix2feng:


contracts/core/modules/lib/TokenFlush.sol
```diff
@@ -0,0 +1,184 @@
...
+
+
+/**
+ * @title TokenFlush
+ * @author Set Protocol
+ *
+ * The TokenFlush contains utility functions to send tokens and base SetTokens from the
+ * Vault or Contract to a specified user address
+ */
+contract TokenFlush is
```
asoong: Is this meant to be reused elsewhere? It's named like it's general but it seems highly coupled with the system. Would prefer it to be specific here instead if it's only done so we can test the library methods. `RebalancingSetExchangeTokenChangeHandler` or something

felix2feng: We previously had it in `RebalancingSetIssuance` - but I realize that we will be using flushing functions for other modules in the future (e.g. token transformation pipeline). So that's why it's refactored to be generic.

asoong: sg!

PR #506, opened 2019-09-05
TimeLockUpgradeV2

felix2feng:
Includes removing the hash and requiring the caller is the owner

### DIFF
--- a/contracts/lib/TimeLockUpgradeV2.sol
+++ b/contracts/lib/TimeLockUpgradeV2.sol
@@ -1,0 +1,152 @@
+/*
+    Copyright 2018 Set Labs Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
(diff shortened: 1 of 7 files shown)


contracts/lib/TimeLockUpgradeV2.sol
```diff
@@ -0,0 +1,152 @@
...
+        uint256 _timestamp
+    );
+
+    event UpgradeUnregistered(
+        bytes32 _upgradeHash
+    );
+
+    /* ============ Modifiers ============ */
+
+    modifier timeLockUpgrade() {
```
bweick: In the future when we make a V2 of something can the first commit just be copying the V1 into the V2 file so then we can have a record of changes after that? Makes much easier to see what the code changes were from V1 to V2.

package.json
```diff
@@ -98,6 +98,7 @@
     "openzeppelin-solidity": "^2.2",
     "set-protocol-utils": "^1.0.0-beta.32",
     "tiny-promisify": "^1.0.0",
+    "truffle-flattener": "^1.4.0",
```
bweick: Not sure you meant to have this in here but I'm ok having it part of our packages since we use it with some regularity

contracts/lib/TimeLockUpgradeV2.sol
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /* ============ Function ============ */
+
+    /**
+     * Removes an existing upgrade.
+     *
+     * @param  _upgradeHash    Keccack256 hash that uniquely identifies function called and arguments 
+     */
+    function deregisterUpgrade(
```
bweick: We should settle on unregister (used in event) vs deregister (used here). Would also be fine with `removeRegisteredUpgrade` or `removeUpgrade` as function name and UpgradeRemoved as event. Deregister and Unregister are weird words. Either way...consistency.

PR #528, opened 2019-11-19
Rebalancing Set V2 Incentives

felix2feng:


contracts/core/tokens/rebalancing-v2/SettleRebalance.sol
```diff
@@ -118,6 +127,42 @@ contract SettleRebalance is
...
     }
 
+    /**
+     * Calculates the fee and mints the rebalancing SetToken quantity to the recipient.
+     * In effect, the Set holders are paying the fee via inflation.
+     */
+    function handleFees()
+        internal
+    {
+        uint256 settlementFee = calculateRebalanceFeeInflation();
```
felix2feng: Question: Do we want to enforce that the fee size is a multiple of the natural unit?

contracts/core/tokens/rebalancing-v2/Issuance.sol
```diff
@@ -82,4 +93,29 @@ contract Issuance is
...
+     */
+    function handleEntryFees(
+        uint256 _quantity
+    )
+        internal
+        returns(uint256)
+    {
+        // The entryFee is a scaled decimal figure by 10e18. We multiply the fee by the quantity
+        // Then descale by 10e18
+        uint256 fee = _quantity.mul(entryFee).deScale();
```
felix2feng: Same here. Does the fee size need to be a multiple of the natural unit?

contracts/mocks/lib/CommonMathMock.sol
```diff
@@ -24,6 +24,26 @@ contract CommonMathMock {
         return CommonMath.safePower(a, pow);
     }
 
+    function testScale(
```
felix2feng: Question: What are your views for using this `scale` pattern to add additional precision when doing math?

PR #531, opened 2019-12-06
Added interface for interacting with RebalancingSetTokenV2.

bweick:


contracts/core/interfaces/IRebalancingSetTokenV2.sol
```diff
@@ -0,0 +1,185 @@
+/*
+    Copyright 2018 Set Labs Inc.
```
felix2feng: 2019

contracts/core/interfaces/IRebalancingSetTokenV2.sol
```diff
@@ -0,0 +1,185 @@
...
+     * @param _nextSet                      The Set to rebalance into
+     *
+     * Can only be called if the rebalance interval has elapsed.
+     * Can only be called by manager.
+     */
+    function startRebalance(
+        address _nextSet
+    )
+        external;
+
```
felix2feng: Let's add settleRebalance

PR #535, opened 2019-12-12
Added ability for dynamic pricePrecision seeking when setting up auct…

bweick:
…ions.


contracts/core/liquidators/impl/Auction.sol
```diff
@@ -205,7 +242,7 @@ contract Auction {
         uint256 currentSetNaturalUnit = _currentSet.naturalUnit();
         uint256 nextNaturalUnit = _nextSet.naturalUnit();
         return Math.max(currentSetNaturalUnit, nextNaturalUnit)
-            .mul(pricePrecision);
+            .mul(_pricePrecision);
     }
 
     /**
```
felix2feng: Do we still need the definition of price to be based on a numerator and denominator if we're removing the ExponentialPivot?

test/contracts/core/liquidators/impl/auction.spec.ts
```diff
@@ -229,18 +275,48 @@ contract('Auction', accounts => {
...
+        await subject();
+
+        const auctionSetup: any = await auctionMock.auction.callAsync();
+
+        const expectedPricePrecision = await liquidatorHelper.calculatePricePrecisionAsync(
+          await coreHelper.getSetInstance(subjectCurrentSet),
+          await coreHelper.getSetInstance(subjectNextSet),
+          oracleWhiteList
+        );
+        console.log(expectedPricePrecision);
```
felix2feng: log

contracts/core/liquidators/impl/Auction.sol
bweick: Yes decaying the denominator was just a way to make it go exponential. But it's still relevant because a numerator of 450 could mean a price equal to .45 (with denominator 1,000) or price of .045 (with a denominator of 1,0000)

test/contracts/core/liquidators/impl/auction.spec.ts
bweick: +1

PR #537, opened 2019-12-16
[V2] Add liquidatorData startRebalance argument

felix2feng:


contracts/core/liquidators/LinearAuctionLiquidator.sol
```diff
@@ -87,15 +87,19 @@ contract LinearAuctionLiquidator is LinearAuction, ILiquidator {
...
         ISetToken _currentSet,
         ISetToken _nextSet,
-        uint256 _startingCurrentSetQuantity
+        uint256 _startingCurrentSetQuantity,
+        bytes calldata _liquidatorData
     )
         external
         isValidSet
     {
+        _liquidatorData; // Pass linting
```
bweick: maybe we create a `handleLiquidatorData` function that this passes to? And then it doesn't nothing with it there. This feels kinda weird, having it in it's own function makes it feel more explicit about what's going on. Not super opinionated though.

bweick:
As a general matter, can you bump the package in this PR then publish it. Imma need this for the SocialTradingManager.

PR #547, opened 2019-12-23
Improved RebalancingSetTokenV2 Logs

felix2feng:
- To save on bytecode size, truncated log size


contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
```diff
@@ -151,6 +151,27 @@ contract RebalancingSetState {
         address oldFeeRecipient
     );
 
+    event EntryFeePaid(
```
felix2feng: @asoong look here

PR #554, opened 2019-12-29
[RBSetV2] Add ability to change entry fee

felix2feng:
- Refactors common assertions into single formulas
- Add ability to change entry fee
- Adds the following state variables: `auctionLibrary`, `proposalPeriod`, `proposalStartTime`

### DIFF
--- a/contracts/core/fee-calculators/FixedFeeCalculator.sol
+++ b/contracts/core/fee-calculators/FixedFeeCalculator.sol
@@ -35,8 +35,6 @@ import { ScaleValidations } from "../../lib/ScaleValidations.sol";
 contract FixedFeeCalculator is IFeeCalculator {
     using SafeMath for uint256;
 
-    /* ============ Constants ============ */
-
     /* ============ State Variables ============ */
 
     // Mapping between an address and its initialization state
--- a/contracts/core/fee-calculators/FixedFeeCalculator.sol
+++ b/contracts/core/fee-calculators/FixedFeeCalculator.sol
@@ -22,6 +22,7 @@ import { CommonMath } from "../../lib/CommonMath.sol";
 import { ICore } from "../interfaces/ICore.sol";
 import { IFeeCalculator } from "../interfaces/IFeeCalculator.sol";
 import { SetTokenLibrary } from "../lib/SetTokenLibrary.sol";
+import { ScaleValidations } from "../../lib/ScaleValidations.sol";
 
 
 /**
@@ -35,8 +36,6 @@ contract FixedFeeCalculator is IFeeCalculator {
     using SafeMath for uint256;
 
     /* ============ Constants ============ */
-
-    uint256 public constant ONE_BASIS_POINT = 10 ** 14;
 
     /* ============ State Variables ============ */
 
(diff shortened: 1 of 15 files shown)


contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
```diff
@@ -187,6 +193,20 @@ contract RebalancingSetState {
         manager = _newManager;
     }
 
+    function setEntryFee(
```
bweick: Can you add this to IRebalancingSetTokenV2 as well

contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
```diff
@@ -127,10 +131,7 @@ contract RebalancingSetState {
     /* ============ Modifier ============ */
 
     modifier onlyManager() {
-        require(
-            msg.sender == manager,
-            "Must be manager"
-        );
+        validateManager();
```
bweick: Are we ever using this in a non-modifier way? Don't know why it needs to be a separate function

felix2feng: It's more bytecode-efficient to separate the function out, as each piece of logic and the reason strings are duplicated across each function that utilizes this modifier

contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
felix2feng: sure

contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
bweick: I see, ok that makes sense

contracts/core/fee-calculators/FixedFeeCalculator.sol
```diff
@@ -36,8 +37,6 @@ contract FixedFeeCalculator is IFeeCalculator {
 
     /* ============ Constants ============ */
```
bweick: Don't need a constants section anymore

contracts/lib/ScaleValidations.sol
```diff
@@ -0,0 +1,39 @@
...
+
+pragma solidity 0.5.7;
+
+import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+library ScaleValidations {
+    using SafeMath for uint256;
+
+    uint256 private constant ONE_HUNDRED_PERCENT = 10 ** 18;
```
bweick: One thing I like the ABDK guys suggested was to start using `1e14`. Didn't realize that was a pattern you could use in Solidity.

contracts/lib/ScaleValidations.sol
```diff
@@ -0,0 +1,39 @@
...
+    uint256 private constant ONE_BASIS_POINT = 10 ** 14;
+
+    function validateLessThanEqualOneHundredPercent(uint256 _value) internal view {
+        require(_value <= ONE_HUNDRED_PERCENT, "Must be < 100%");
```
bweick: <= in msg

contracts/core/tokens/rebalancing-v2/RebalancingSetState.sol
```diff
@@ -17,6 +17,8 @@
 pragma solidity 0.5.7;
 pragma experimental "ABIEncoderV2";
 
+import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
bweick: Where is `SafeMath` being used?

PR #565, opened 2020-01-08
Rebalancing Set cToken Bidding Contract

richardliang:


contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,360 @@
...
+    // Address and instance of CToken contract
+    ICToken public targetCToken;
+
+    // Address of underlying of the target CToken contract
+    address public underlyingAddress;
+
+    string public dataDescription;
+
+    // Exchange Rate values are scaled by 1e18
+    uint256 internal constant scalingFactor = 10 ** 18;
```
felix2feng: You can actually use `scaleFactor` in commonMath instead of redeclaring here

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,360 @@
...
+    );
+
+    /* ============ Constructor ============ */
+
+    /**
+     * Constructor function for RebalancingSetCTokenBidder
+     *
+     * @param _rebalanceAuctionModule   The address of RebalanceAuctionModule
+     * @param _transferProxy            The address of TransferProxy
+     * @param _targetCToken                   The address of the target cToken
```
felix2feng: So this only works for a particular cToken? Each cToken needs its own bidder?

felix2feng: Also comment spacing

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,360 @@
...
+    constructor(
+        IRebalanceAuctionModule _rebalanceAuctionModule,
+        ITransferProxy _transferProxy,
+        ICToken _targetCToken,
+        address _underlyingAddress,
+        string memory _dataDescription
+    )
+        public
+    {
+        // Commit the RebalanceAuctionModule instance
```
felix2feng: With the constructor, I'm okay not adding a comment for each assignment

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,360 @@
...
+    using SafeMath for uint256;
+    
+    // Address and instance of RebalanceAuctionModule contract
+    IRebalanceAuctionModule public rebalanceAuctionModule;
+
+    // Address and instance of TransferProxy contract
+    ITransferProxy public transferProxy;
+
+    // Address and instance of CToken contract
+    ICToken public targetCToken;
```
felix2feng: If we want to reuse this contract, we can create a simple mapping of underlying -> cToken.

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,360 @@
...
+                // Calculated as cToken quantity * exchangeRate / scalingFactor.
+                uint256 exchangeRate = targetCToken.exchangeRateStored();
+                uint256 currentInflowQuantity = inflowArray[i];
+                uint256 currentOutflowQuantity = outflowArray[i];
+
+                inflowArray[i] = currentInflowQuantity.mul(exchangeRate).div(scalingFactor);
+                outflowArray[i] = currentOutflowQuantity.mul(exchangeRate).div(scalingFactor);
+
+                // Check for rounding error and add 1 if needed
+                inflowArray[i] = inflowArray[i].mul(scalingFactor).div(exchangeRate) >= currentInflowQuantity
```
felix2feng: In what cases would there be a rounding error? 

There may be a better way to handle this.

contracts/helper/RebalancingSetCTokenBidder.sol
richardliang: Yeah, since imagine that we would create a v2 when we do support cDAI in social trading. And v3 when we may include cETH support and ability to bid in ETH like in the ETH bidder

But now do think a simple mapping for all the existing cTokens minus cETH makes sense

contracts/helper/RebalancingSetCTokenBidder.sol
richardliang: Seems like when converting from cToken amount to underlying. Since required underlying amount is less than cToken amount, and if its an odd number

I copied the logic from instadapp https://github.com/InstaDApp/smart-contract/blob/master/contracts/bridges/MakerCompound.sol#L394

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+        );
+    }
+
+    /*
+     * Get token inflows and outflows and combined token array denominated in underlying required
+     * for bid for a given rebalancing Set token.
+     *
+     * @param _quantity               The amount of currentSet to be rebalanced
+     * @param _rebalancingSetToken    The rebalancing Set Token instance
+     * @return inflowArray            Array of amount of tokens inserted into system in bid
```
felix2feng: Funnily, we actually have a term for these outputs. It's called `TokenFlow`. You can find in the `Rebalance.sol` file. That said, i think bid price and address array is clearer for third parties.

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+     * @return outflowArray           Array of amount of tokens returned from system in bid
+     * @return combinedTokenArray     Array of token addresses
+     */
+    function getBidPriceAndAddressArray(
+        uint256 _quantity,
```
felix2feng: Typical argument order is the token, then the quantity

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+        IRebalancingSetToken _rebalancingSetToken
+    )
+        public
+        view
+        returns (uint256[] memory, uint256[] memory, address[] memory)
```
felix2feng: Typical order is addresses then quantities

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+    {
+        // Get token addresses
+        address[] memory combinedTokenArray = _rebalancingSetToken.getCombinedTokenArray();
```
felix2feng: To reduce duplication in this method and the above, you can create an internal function called `getTokenFlows` that returns the addresses + inflow./outflow

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+                // Calculated as cToken quantity * exchangeRate / 10 ** 18.
+                uint256 exchangeRate = ICToken(currentComponent).exchangeRateStored();
+                uint256 currentInflowQuantity = inflowArray[i];
+                uint256 currentOutflowQuantity = outflowArray[i];
+
+                inflowArray[i] = currentInflowQuantity.mul(exchangeRate).div(CommonMath.scaleFactor());
+                outflowArray[i] = currentOutflowQuantity.mul(exchangeRate).div(CommonMath.scaleFactor());
+
+                // Check for rounding error and add 1 if needed
+                inflowArray[i] = inflowArray[i].mul(CommonMath.scaleFactor()).div(exchangeRate) >= currentInflowQuantity
```
felix2feng: To simplify this, use `CommonMath. divCeil` instead - which I think achieves what you want? You can even do it in the above function.

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+                    currentComponentQuantity
+                );
+
+                // If cToken, calculate required underlying tokens, transfer to contract, 
+                // ensure underlying allowance to cToken and then mint cTokens
+                if (cTokenToUnderlying[currentComponent] != address(0)) {
+                    address underlyingAddress = cTokenToUnderlying[currentComponent];
+
+                    // Calculate required amount of underlying. Calculated as cToken quantity * exchangeRate / 10 ** 18.
+                    uint256 exchangeRate = ICToken(currentComponent).exchangeRateCurrent();
```
felix2feng: Since this is performed multiple times, you should make the calculation of underlying a private function. In fact, I'd create a cToken library (since I'd probably need to use this in the piping)

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+     * @param _transferProxy            The address of TransferProxy
+     * @param _cTokenArray              The address array of the target cToken
+     * @param _underlyingArray          The address array of the target cToken's underlying
+     * @param _dataDescription          Description of contract for Etherscan / other applications
+     */
+    constructor(
+        IRebalanceAuctionModule _rebalanceAuctionModule,
+        ITransferProxy _transferProxy,
+        address[] memory _cTokenArray,
+        address[] memory _underlyingArray,
```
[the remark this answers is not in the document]
richardliang: Thinking that because this isn't a core part of the system, we can make upgrades and redeploy instead, similar to ProtocolViewer. Imagine we do want a function that transacts in ETH in a future version so we can implement adding/removing governance as well

contracts/helper/RebalancingSetCTokenBidder.sol
richardliang: +1

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,357 @@
...
+    using SafeMath for uint256;
+    
+    // Address and instance of RebalanceAuctionModule contract
+    IRebalanceAuctionModule public rebalanceAuctionModule;
+
+    // Address and instance of TransferProxy contract
+    ITransferProxy public transferProxy;
+
+    // Mapping of cToken address to underlying address
+    mapping (address => address) public cTokenToUnderlying;
```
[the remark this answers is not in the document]
richardliang: The reason I didn't use IERC20 was cause we didnt need access to token functions, and for both ICToken and IERC20 we have to convert the instance to address a lot for checks i.e.`cTokenToUnderlying[currentComponent] != address(0)`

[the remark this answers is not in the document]
richardliang: Dont think you can have ICToken as a key value actually

ParserError: Expected elementary type name for mapping key type
    mapping (ICToken => address) public cTokenToUnderlying;

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,369 @@
...
+    )
+        external
+        nonReentrant
+    {
+        /// Get token flow arrays for the given bid quantity
+        address[] memory combinedTokenArray;
+        uint256[] memory inflowArray;
+        uint256[] memory outflowArray;
+        (
+            combinedTokenArray,
```
felix2feng: Can make more efficient by doing the following
```
(
   address[] memory combinedTokenArray,
  ...
```

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,369 @@
...
+     * @param _quantity               The amount of currentSet to be rebalanced
+     * @return combinedTokenArray     Array of token addresses
+     * @return inflowArray            Array of amount of tokens inserted into system in bid
+     * @return outflowArray           Array of amount of tokens returned from system in bid
+     */
+    function getAddressAndBidPriceArray(
+        IRebalancingSetToken _rebalancingSetToken,
+        uint256 _quantity
+    )
+        public
```
felix2feng: Function modifier can be external if we don't expect it to be inherited

contracts/lib/CompoundUtils.sol
```diff
@@ -0,0 +1,48 @@
...
+ * @title CompoundUtils
+ * @author Set Protocol
+ *
+ * Collection of common Compound functions for use in Set Protocol contracts
+ */
+library CompoundUtils
+{
+    using SafeMath for uint256;
+
+    function calculateUnderlyingUnits(
```
felix2feng: Suggestion: name could be convertcTokenToUnderlying

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Get token flows array of addresses, inflows and outflows
+     *
+     * @param    _rebalancingSetToken   The rebalancing Set Token instance
+     * @param    _quantity              The amount of currentSet to be rebalanced
+     * @return   combinedTokenArray     Array of token addresses
+     * @return   inflowArray            Array of amount of tokens inserted into system in bid
+     * @return   outflowArray           Array of amount of tokens returned from system in bid
+     */
+    function getTokenFlows(
```
felix2feng: This function could potentially be placed in `Rebalance.sol` and reused.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "set-protocol-contracts",
-  "version": "1.3.18-beta",
+  "version": "1.3.15-beta",
```
felix2feng: Version going down?

contracts/helper/RebalancingSetCTokenBidder.sol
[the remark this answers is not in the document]
bweick: Does Compound of a central repository that links underlying to cToken contract? Maybe we could just use something like that if it exists

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -0,0 +1,369 @@
...
+        public
+        view
+        returns (address[] memory, uint256[] memory, uint256[] memory)
+    {
+        // Get token flow arrays for the given bid quantity
+        address[] memory combinedTokenArray;
+        uint256[] memory inflowArray;
+        uint256[] memory outflowArray;
+        (
+            combinedTokenArray,
```
bweick: Same here as Felix pointed out above

PR #574, opened 2020-01-28
RebalancingSetToken V3 with Generalized Incentive Fees

felix2feng:


contracts/core/tokens/RebalancingSetTokenV3.sol
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * Anyone can call this function.
+     */
+    function actualizeFee()
+        external
+    {
+        IncentiveFee.validateFeeActualization();
+
+        // Calculates fees and mints Rebalancing Set to the feeRecipient, increasing supply
+        (uint256 feePercent, uint256 feeQuantity) = RebalancingSettlement.handleFees();
```
bweick: We might have an issue here because IFeeCalculator `getFee` is a `view` function but now alters state in order to set high watermarks and timestamps.

felix2feng: Ok i see. Let's make it non-view.

contracts/core/tokens/rebalancing-v3/IncentiveFee.sol
```diff
@@ -0,0 +1,123 @@
...
+     * The formula is as follows:
+     * newUnitShares = currentSetAmount * rebalanceSetNaturalUnit / rebalanceSetTotalSupply
+     */
+    function calculateNewUnitShares() internal view returns(uint256) {
+        uint256 currentSetAmount = vault.getOwnerBalance(
+            address(currentSet),
+            address(this)
+        );
+
+        return currentSetAmount.mul(naturalUnit).div(totalSupply());
```
bweick: Thinking the we might need to round up the math to determine the amount of naturalUnits that have been issued (i.e. `ceil(totalSupply.div(naturalUnit))`). Reason being that if we round down in theory (though perhaps not in practice due to redeem restrictions of the naturalUnit) we would have an under-collateralized Set. Let's say we have 5.5 naturalUnits worth of the RebalancingSet and 500 units of the collateralSet, the unitShares would be equal to 100. But in theory there would be a claim to 550 collateralized Sets, so it would be under-collateralized. If we round up unitShares is 83.3333, which would lead to a claim on 458.33 collateralSets.

contracts/core/tokens/RebalancingSetTokenV3.sol
```diff
@@ -0,0 +1,161 @@
...
+
+    /*
+     * Overrides the RebalancingSetTokenV2 settleRebalance function.
+     *
+     * After a successful rebalance, the new Set is issued. 
+     * Full issuance functionality is now returned to set owners. No fees are captured.
+     *
+     * Anyone can call this function.
+     */
+    function settleRebalance()
```
bweick: Do we want to take advantage of settlement to also accrue fees?

contracts/core/fee-calculators/FixedFeeCalculator.sol
```diff
@@ -79,6 +79,16 @@ contract FixedFeeCalculator is IFeeCalculator {
         return fees[msg.sender];
     }
 
+    /**
```
bweick: Do we want to version this file `FixedFeeCalculatorV2.sol` so that it's clear this is an updated version from the original one we deployed?

contracts/core/tokens/rebalancing-v3/IncentiveFee.sol
```diff
@@ -0,0 +1,123 @@
...
+ */
+contract IncentiveFee is 
+    ERC20,
+    RebalancingSetState
+{
+    using SafeMath for uint256;
+
+    /* ============ Events ============ */
+
+    event IncentiveFeePaid(
```
bweick: Do we need the rebalancingSet address in here too? What's easiest for Alex to handle?

contracts/core/tokens/rebalancing-v3/IncentiveFee.sol
```diff
@@ -0,0 +1,123 @@
...
+    function validateFeeActualization() internal view {
+        validateRebalanceStateIs(RebalancingLibrary.State.Default);
+    }
+
+    /*
+     * After the minting of new inflation fees, the unit shares must be updated.
+     * The formula is as follows:
+     * newUnitShares = currentSetAmount * rebalanceSetNaturalUnit / rebalanceSetTotalSupply
+     */
+    function calculateNewUnitShares() internal view returns(uint256) {
```
bweick: Can't we make this a general function that can be used here and in settlement. You would have to pass in the collateralSet address but that seems fine.

contracts/core/tokens/rebalancing-v3/IncentiveFee.sol
felix2feng: In this case no - because the Set itself is emitting the event.

contracts/core/tokens/rebalancing-v3/IncentiveFee.sol
felix2feng: +1

contracts/core/fee-calculators/FixedFeeCalculator.sol
felix2feng: Currently leaning towards leaving it - as deployed versions are not currently utilized and we plan not to going forward

PR #575, opened 2020-01-31
[WIP] Performance Fee Calculator

bweick:
…e based off a high watermark. The high watermark can be reset after a period of time. Some refactoring done of helpers.

### DIFF
--- a/contracts/core/fee-calculators/PerformanceFeeCalculator.sol
+++ b/contracts/core/fee-calculators/PerformanceFeeCalculator.sol
@@ -46,7 +46,8 @@ contract PerformanceFeeCalculator is IFeeCalculator {
     event FeeActualized(
         address indexed rebalancingSetToken,
         uint256 newHighWatermark,
-        uint256 inflationFee
+        uint256 profitFee,
+        uint256 streamingFee
     );
 
     /* ============ Structs ============ */
@@ -170,7 +171,7 @@ contract PerformanceFeeCalculator is IFeeCalculator {
     {
         require(
             core.validSets(msg.sender),
-            "PerformanceFeeCalculator.getFee: Caller must be valid RebalancingSetToken."
+            "PerformanceFeeCalculator.updateAndGetFee: Caller must be valid RebalancingSetToken."
         );
 
         uint256 streamingFee = calculateStreamingFee();
@@ -183,7 +184,8 @@ contract PerformanceFeeCalculator is IFeeCalculator {
         emit FeeActualized(
             msg.sender,
             highWatermark(msg.sender),
-            streamingFee.add(profitFee)
+            profitFee,
+            streamingFee
         );
 
         return streamingFee.add(profitFee);
@@ -260,7 +262,7 @@ contract PerformanceFeeCalculator is IFeeCalculator {
     function validateFeeParameters(
         InitFeeParameters memory parameters
     )
-        private
+        internal
         view
     {
         require(
--- a/contracts/core/fee-calculators/PerformanceFeeCalculator.sol
+++ b/contracts/core/fee-calculators/PerformanceFeeCalculator.sol
@@ -1,0 +1,291 @@
+/*
+    Copyright 2019 Set Labs Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+pragma solidity 0.5.7;
+
+import { SafeMath } from "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+import { CommonMath } from "../../lib/CommonMath.sol";
+import { ICore } from "../interfaces/ICore.sol";
+import { IFeeCalculator } from "../interfaces/IFeeCalculator.sol";
+import { IOracleWhiteList } from "../interfaces/IOracleWhiteList.sol";
+import { IRebalancingSetTokenV2 } from "../interfaces/IRebalancingSetTokenV2.sol";
+import { ISetToken } from "../interfaces/ISetToken.sol";
+import { ScaleValidations } from "../../lib/ScaleValidations.sol";
+import { SetUSDValuation } from "../liquidators/impl/SetUSDValuation.sol";
+
+
+/**
+ * @title PerformanceFeeCalculator
+ * @author Set Protocol
+ *
+ * Smart contract that stores and returns fees (represented as scaled decimal values). Fees are
+ * determined based on performance of the Set and a streaming fee.
+ */
+contract PerformanceFeeCalculator is IFeeCalculator {
+
+    using SafeMath for uint256;
+    using CommonMath for uint256;
+
+    /* ============ Events ============ */
+
+    event FeeAcutalized(
+        address indexed rebalancingSetToken,
+        uint256 oldHighWatermark,
+        uint256 newHighWatermark,
+        uint256 inflationFee
+    );
+
+    /* ============ Structs ============ */
+    struct FeeState {
+        uint256 profitFeeFrequency;
+        uint256 highWatermarkResetFrequency;
+        uint256 profitFeePercentage;
+        uint256 streamingFeePercentage;
+        uint256 highWatermark;
+        uint256 lastProfitFeeTimestamp;
+        uint256 lastStreamingFeeTimestamp;
+    }
+
+    struct InitFeeParameters {
+        uint256 profitFeeFrequency;
+        uint256 highWatermarkResetFrequency;
+        uint256 profitFeePercentage;
+        uint256 streamingFeePercentage;        
+    }
+
+    /* ============ Constants ============ */
+    // 365.25 days used to represent the year
+    uint256 private constant ONE_YEAR_IN_SECONDS = 31557600;
+    uint256 private constant ONE_HUNDRED_PERCENT = 1e18;
+
+    /* ============ State Variables ============ */
+    ICore public core;
+    IOracleWhiteList public oracleWhiteList;
+    uint256 public maximumProfitFeePercentage;
+    uint256 public maximumStreamingFeePercentage;
+    mapping(address => FeeState) public feeState;
+
+    /* ============ Constructor ============ */
+
+    constructor(
+        ICore _core,
+        IOracleWhiteList _oracleWhiteList,
+        uint256 _maximumProfitFeePercentage,
+        uint256 _maximumStreamingFeePercentage
+    )
+        public
+    {
+        core = _core;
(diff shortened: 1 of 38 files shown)


contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,312 @@
...
+            feeState[msg.sender].highWatermark = postStreamingValue;
+        }
+    }
+
+    function exceedsProfitFeeFrequency() internal view returns (bool) {
+        return block.timestamp.sub(lastProfitFeeTimestamp()) > profitFeeFrequency();
+    }
+
+    function lastStreamingFeeTimestamp() internal view returns (uint256) {
+        return feeState[msg.sender].lastStreamingFeeTimestamp;
```
felix2feng: For more flexibility and potential reusability, I prefer passing in the setAddress for each of these getters. 

The main usecase is that we may want to expose external getter functions to retrieve these figures.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,312 @@
...
+
+        return timeSinceLastFee.mul(streamingFeePercentage()).div(ONE_YEAR_IN_SECONDS);
+    }
+
+    function calculateProfitFee(uint256 _streamingFee)
+        internal
+        view
+        returns(uint256)
+    {
+        uint256 rebalancingSetValue = calculateRebalancingSetValue(msg.sender);
```
felix2feng: Random note - I would pass the Set address / msg.sender into this function to enable further flexibility

bweick: ? what are you referencing?

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
+/*
+    Copyright 2019 Set Labs Inc.
```
felix2feng: 2020

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+ * in any any asset based on oracle white list used in deploy.
+ */
+contract PerformanceFeeCalculator is IFeeCalculator {
+
+    using SafeMath for uint256;
+    using CommonMath for uint256;
+
+    /* ============ Events ============ */
+
+    event FeeAcutalized(
```
felix2feng: `FeeActualized`

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /* ============ External Functions ============ */
+
+    /*
+     * Called by RebalancingSetToken, parses bytedata then assigns to correct FeeState struct. 
+     *
+     * @param  _feeCalculatorData       Bytestring encoding fee parameters for RebalancingSetToken
+     */
+    function initialize(
```
felix2feng: Right now the creation flow is as follows:

Set gets created -> Factory calls initialize on the Set -> Set calls FeeCalculator.

This happens all in one transaction. I don't see how someone else would be able to front-run or anything. Either 1) the Set does not exist yet or 2) the caller is not the Set.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        return streamingFee.add(profitFee);
+    }
+
+    /*
+     * Calculates total inflation percentage in order to accrue fees to manager. Profit fee calculations
+     * are net of streaming fees, so streaming fees are applied first then profit fees are calculated. 
+     * Additionally, fee state is set timestamps are updated for each fee type and the high watermark is
+     * reset if time since last profit fee exceeds the highWatermarkResetFrequency. 
+     *
+     * @param  uint256       Percent inflation of supply
```
felix2feng: @return

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        feeState[msg.sender].lastProfitFeeTimestamp = block.timestamp;
+        feeState[msg.sender].lastStreamingFeeTimestamp = block.timestamp;
+        feeState[msg.sender].highWatermark = calculateRebalancingSetValue(msg.sender);
+    }
+
+    /*
+     * Calculates total inflation percentage in order to accrue fees to manager. Profit fee calculations
+     * are net of streaming fees, so streaming fees are applied first then profit fees are calculated.  
+     *
+     * @param  uint256       Percent inflation of supply
```
felix2feng: @return

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+     */
+    function updateAndGetFee()
+        external
+        returns (uint256)
+    {
+        uint256 streamingFee = calculateStreamingFee();
+
+        uint256 profitFee = 0;
```
felix2feng: Could deduplicate this code by creating an internal function called `getProfitFee` - which returns the value based on this logic here.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        // If profit fee then set new high watermark and profit fee timestamp
+        if (_profitFee > 0) {
+            feeState[msg.sender].lastProfitFeeTimestamp = block.timestamp;
+            feeState[msg.sender].highWatermark = postStreamingValue;
+            return;
+        }
+
+        // If no profit fee and last profit fee was more than highWatermarkResetFrequency seconds ago then reset
+        // high watermark
+        if (block.timestamp.sub(lastProfitFeeTimestamp(msg.sender)) >= highWatermarkResetFrequency(msg.sender)) {
```
felix2feng: To make it more readable, could use an `else if ` instead of return

felix2feng: Would prefer a variable called `timeSinceLastProfitFee` if possible

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        );
+    }
+
+    /**
+     * Calculates streaming fee by multiplying streamingFeePercentage by the elapsed amount of time since the last fee
+     * was collected divided by one year in seconds, since the fee is a yearly fee.
+     *
+     * @return uint256        Streaming fee
+     */
+    function calculateStreamingFee()
```
felix2feng: Wondering if we would want the resulting fee to be a multiple of a basis point for any reason.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
felix2feng: Sorry, I would make the rebalancingSet a parameter to this function for extra flexibility.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        // Determine percent fee in terms of current rebalancing Set value
+        return gainedValue.mul(profitFeePercentage(msg.sender)).div(rebalancingSetValue);
+    }
+
+    /**
+     * Calculates value of RebalancingSetToken.
+     *
+     * @return uint256        Streaming fee
+     */
+    function calculateRebalancingSetValue(
```
felix2feng: Since, I feel this would be reused a lot, we should move this into its own library. (Would need to pass in the oracleWhitelist as a parameter if so)

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,383 @@
...
+        );
+
+        // Value of rebalancing set is collateralValue times unitShares divided by naturalUnit
+        return collateralValue.mul(unitShares).div(naturalUnit);
+    }
+
+    /**
+     * Checks if time since last profit fee exceeds profitFeeFrequency
+     */
+    function exceedsProfitFeeFrequency() internal view returns (bool) {
```
felix2feng: To confirm to pattern below, would pass in the `address _set` as well here

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
bweick: Isn't it already? What function are you talking about?

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
bweick: Can't think of one in this instance, think thats more of a product thing that we do than any security related reasons.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,377 @@
...
+     *
+     * @preturn  uint256       Percent inflation of supply
+     */
+    function updateAndGetFee()
+        external
+        returns (uint256)
+    {
+        require(
+            core.validSets(msg.sender),
+            "PerformanceFeeCalculator.getFee: Caller must be valid RebalancingSetToken."
```
felix2feng: `PerformanceFeeCalculator .updateAndGetFee`

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,377 @@
...
+
+    using SafeMath for uint256;
+    using CommonMath for uint256;
+
+    /* ============ Events ============ */
+
+    event FeeActualized(
+        address indexed rebalancingSetToken,
+        uint256 newHighWatermark,
+        uint256 inflationFee
```
felix2feng: In addition to inflationFee, I think it would be useful to print out what the `streamingFee` and `profitFee`. How else would the backend know what to use? Or otherwise, it would need to be implied.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,377 @@
...
+
+        uint256 profitFee = calculateProfitFee(streamingFee);
+
+        // Update fee state based off fees collected
+        updateFeeState(streamingFee, profitFee);
+
+        emit FeeActualized(
+            msg.sender,
+            highWatermark(msg.sender),
+            streamingFee.add(profitFee)
```
felix2feng: Consider putting it into a local variable

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    /*
+     * Validates fee parameters. Ensures that both fees are below the max fee percentages and that they are
+     * multiples of a basis point. Also makes sure highWatermarkResetFrequency is greater than profitFeeFrequency.
+     */
+    function validateFeeParameters(
+        InitFeeParameters memory parameters
+    )
+        private
```
felix2feng: Let's make it internal - as maybe we'd want to overwrite it in a future iteration.

PR #577, opened 2020-02-04
Generalize to AddressToAddressWhitelist from OracleWhiteList

richardliang:
- Renamed var and descriptions

### DIFF
--- a/artifacts/index.ts
+++ b/artifacts/index.ts
@@ -69,7 +69,6 @@ export {
   ExchangeIssuanceLibraryContract,
   ExchangeIssuanceModuleContract,
   FailAuctionLibraryContract,
-  FactoryUtilsLibraryContract,
   FixedFeeCalculatorContract,
   IAuctionPriceCurveContract,
   InvalidReturnTokenMockContract,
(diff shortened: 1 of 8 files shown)


artifacts/index.ts
```diff
@@ -69,7 +69,6 @@ export {
   ExchangeIssuanceLibraryContract,
   ExchangeIssuanceModuleContract,
   FailAuctionLibraryContract,
-  FactoryUtilsLibraryContract,
```
felix2feng: Any reason to delete?

utils/helpers/coreHelper.ts
```diff
@@ -3,6 +3,7 @@ import * as setProtocolUtils from 'set-protocol-utils';
 import { Address } from 'set-protocol-utils';
 
 import {
+  AddressToAddressWhiteListContract,
```
felix2feng: Since this isn't related to "Core", I would suggest putting this in a separate helper - like libraryHelper or utilsHelper

artifacts/index.ts
richardliang: It doesnt exist anymore? Failed previously when compiling

utils/helpers/coreHelper.ts
richardliang: 👍

PR #579, opened 2020-02-07
cToken Issuance Piping

richardliang:


contracts/core/modules/CTokenExchangeIssuanceModule.sol
```diff
@@ -0,0 +1,439 @@
...
+        // Calculate expected receive token balances
+        uint256[] memory requiredBalances = calculateReceiveTokenBalances(
+            _exchangeIssuanceParams
+        );
+
+        // Execute the exchange orders using the encoded order data and deposits tokens in vault
+        executeExchangeOrders(_orderData);
+
+        // Withdraw underlying tokens required from vault, mint cTokens in the module, and return to vault under sender address
+        mintCTokens(
```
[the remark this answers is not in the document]
felix2feng: Though it may be okay for the short term. What does the gas cost look like for let's say a 100% cUSDC Set? And what if its 50% cUSDC, 50% ETH?

contracts/core/modules/CTokenExchangeIssuanceModule.sol
```diff
@@ -0,0 +1,439 @@
...
+            _exchangeIssuanceParams.receiveTokens,
+            _exchangeIssuanceParams.receiveTokenAmounts
+        );
+
+        // Check that sender's receive tokens in Vault have been incremented correctly
+        ExchangeIssuanceLibrary.validatePostExchangeReceiveTokenBalances(
+            vault,
+            _exchangeIssuanceParams.receiveTokens,
```
felix2feng: In this case, the receiveTokens are the cTokens or the components again?

contracts/core/modules/RebalancingSetCTokenIssuanceModule.sol
```diff
@@ -0,0 +1,593 @@
...
+            _vault,
+            _transferProxy,
+            _weth
+        )
+    {
+        cTokenWhiteList = _cTokenWhiteList;
+
+        address[] memory cTokenAddresses = _cTokenWhiteList.validAddresses();
+
+        for (uint256 i = 0; i < cTokenAddresses.length; i++) {
```
felix2feng: This is the 3rd time this logic has been replicated (cTokenExchangeIssuanceModule, cTokenIssuanceModule, and the bidder. Perhaps, we should just create a file that contract that is inherited that has 1) the cToken Whitelist state and the approval logic in the constructor.

contracts/core/modules/CTokenExchangeIssuanceModule.sol
richardliang: These are cTokens. For redeeming, the sendTokens are the cTokens

contracts/lib/AddressToAddressWhiteList.sol
```diff
@@ -37,19 +37,19 @@ contract AddressToAddressWhiteList is
 
     /* ============ State Variables ============ */
 
-    address[] public addresses;
-    mapping(address => address) public addressToAddressWhiteList;
+    address[] public keys;
+    mapping(address => address) public keysToValues;
```
felix2feng: Looking at this again, we can just call it `whitelist` instead of `keysToValues` lol.

contracts/core/modules/lib/CTokenModuleCoreState.sol
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * The CTokenModuleCoreState library maintains Core-related state for modules that interact with cTokens.
+ *
+ */
+contract CTokenModuleCoreState {
+
+    /* ============ State Variables ============ */
+
+    // Address of transferProxy contract
+    address public transferProxy;
```
felix2feng: This construct is a bit too rigid. Frankly, we should probably have a `CoreModuleState2` that has the `transferProxy` in it. And we could name this something like `cTokenWhitelisted` that only has the cTokenWhitelist and passes in the whitelist

contracts/core/modules/lib/CTokenModuleCoreState.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    // Address and instance of AddressToAddressWhiteList contract
+    IAddressToAddressWhiteList public cTokenWhiteList; 
+
+    /* ============ Public Getters ============ */
```
felix2feng: Constructor

contracts/core/modules/lib/CTokenModuleCoreState.sol
```diff
@@ -0,0 +1,86 @@
...
+    constructor(
+        address _transferProxy,
+        IAddressToAddressWhiteList _cTokenWhiteList
+    )
+        public
+    {
+        transferProxy = _transferProxy;
+        cTokenWhiteList = _cTokenWhiteList;
+
+        address[] memory cTokenAddresses = _cTokenWhiteList.validAddresses();
```
felix2feng: Add a note that we do not expect there to be many validAddresses set, so we won't hit gas limit issues.

contracts/core/modules/RebalancingSetCTokenIssuanceModule.sol
```diff
@@ -0,0 +1,569 @@
...
+        address _vault,
+        address _transferProxy,
+        IWETH _weth,
+        IAddressToAddressWhiteList _cTokenWhiteList
+    )
+        public
+        RebalancingSetIssuanceModule(
+            _core,
+            _vault,
+            _transferProxy,
```
felix2feng: Interesting - transferProxy is set twice?

felix2feng: In this case, transferProxy isn't needed in the `cTokenModuleCoreState`

contracts/lib/AddressToAddressWhiteList.sol
richardliang: +1

PR #595, opened 2020-03-02
Brian/allow fee update

bweick:


contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -208,6 +213,34 @@ contract PerformanceFeeCalculator is IFeeCalculator {
         return streamingFee.add(profitFee);
     }
 
+    /*
+     * Validate then set new streaming fee.
+     *
+     * @param  _newFeePercentage       Fee type and new streaming fee encoded in bytes
+     */
+    function adjustFee(
```
felix2feng: I imagine an event should be emitted here?

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -208,6 +213,34 @@ contract PerformanceFeeCalculator is IFeeCalculator {
...
+            uint256 feePercentage
+        ) = parseNewFeeCallData(_newFeePercentage);
+
+        // Since only two fee options and anything feeType integer passed that is not 0 or 1 will revert can
+        // make this a simple if...else... statement
+        if (feeIdentifier == FeeType.StreamingFee) {
+            validateStreamingFeePercentage(feePercentage);
+
+            feeState[msg.sender].streamingFeePercentage = feePercentage;
+        } else {
```
felix2feng: I feel better if we have explicit 
```
else if (feeIdentifier == 1)
```

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -208,6 +213,34 @@ contract PerformanceFeeCalculator is IFeeCalculator {
...
 
+    /*
+     * Validate then set new streaming fee.
+     *
+     * @param  _newFeePercentage       Fee type and new streaming fee encoded in bytes
+     */
+    function adjustFee(
+        bytes calldata _newFeePercentage
+    )
+        external
```
felix2feng: We need to validate that only the RBSet can call this function

contracts/core/interfaces/IRebalancingSetTokenV3.sol
```diff
@@ -0,0 +1,350 @@
...
+pragma solidity 0.5.7;
+pragma experimental "ABIEncoderV2";
+
+import { RebalancingLibrary } from "../lib/RebalancingLibrary.sol";
+import { IFeeCalculator } from "./IFeeCalculator.sol";
+import { ILiquidator } from "./ILiquidator.sol";
+import { ISetToken } from "./ISetToken.sol";
+
+/**
+ * @title IRebalancingSetTokenV2
```
felix2feng: V2?

contracts/core/tokens/RebalancingSetTokenV3.sol
```diff
@@ -146,8 +146,24 @@ contract RebalancingSetTokenV3 is
         );
     }
 
+    /*
+     * Validate then set new streaming fee.
```
felix2feng: Update me.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
bweick: I would too but then we have coverage issues cause anything that doesn't fall under those two options would have reverted before and coverage will say there's a branch that hasn't been explored.

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
bweick: Why? The only state that will update is the callers, which would be the RebalancingSetToken.

felix2feng: hmm i see, nevermind then

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
bweick: It is on the FeeCalculator

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -208,6 +206,34 @@ contract PerformanceFeeCalculator is IFeeCalculator {
         return streamingFee.add(profitFee);
     }
 
+    /*
+     * Validate then set new streaming fee.
+     *
+     * @param  _newFeePercentage       Fee type and new streaming fee encoded in bytes
```
felix2feng: Variable name isn't that accurate, since its not just the percentage. Maybe it is `newFeeData`

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
felix2feng: We could add a else block that reverts "Invalid feeidentifier"

bweick: That's what I'm saying it won't get to that point cause say you try to assign 3 to the enum it will revert when you're parsing the call data.

bweick: Since there's no switch case in Solidity we either have to be ok with this pattern or skip coverage on this line

PR #596, opened 2020-03-03
Add TimeLockUpgradeV3 that allows for more arbitrary logic. Remove mo…

bweick:
…ckLibraryHelper from exports.

### DIFF
--- a/artifacts/index.ts
+++ b/artifacts/index.ts
@@ -2,23 +2,16 @@ export { AddressToAddressWhiteList } from "./ts/AddressToAddressWhiteList";
 export { AuctionMock } from "./ts/AuctionMock";
 export { BadCTokenMock } from "./ts/BadCTokenMock";
 export { BadTokenMock } from "./ts/BadTokenMock";
-export { Bytes32LibraryMock } from "./ts/Bytes32LibraryMock";
-export { CommonMathMock } from "./ts/CommonMathMock";
 export { CommonValidationsLibrary } from "./ts/CommonValidationsLibrary";
-export { CommonValidationsLibraryMock } from "./ts/CommonValidationsLibraryMock";
-export { CompoundUtilsMock } from "./ts/CompoundUtilsMock";
 export { ConstantAuctionPriceCurve } from "./ts/ConstantAuctionPriceCurve";
 export { Core } from "./ts/Core";
 export { CoreIssuanceLibrary } from "./ts/CoreIssuanceLibrary";
-export { CoreIssuanceLibraryMock } from "./ts/CoreIssuanceLibraryMock";
 export { CoreMock } from "./ts/CoreMock";
 export { CTokenExchangeIssuanceModule } from "./ts/CTokenExchangeIssuanceModule";
 export { ERC20 } from "./ts/ERC20";
 export { ERC20Detailed } from "./ts/ERC20Detailed";
 export { ERC20Wrapper } from "./ts/ERC20Wrapper";
-export { ERC20WrapperMock } from "./ts/ERC20WrapperMock";
 export { ExchangeIssuanceLibrary } from "./ts/ExchangeIssuanceLibrary";
-export { ExchangeIssuanceLibraryMock } from "./ts/ExchangeIssuanceLibraryMock";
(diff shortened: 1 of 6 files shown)


artifacts/index.ts
```diff
@@ -2,23 +2,16 @@ export { AddressToAddressWhiteList } from "./ts/AddressToAddressWhiteList";
 export { AuctionMock } from "./ts/AuctionMock";
 export { BadCTokenMock } from "./ts/BadCTokenMock";
 export { BadTokenMock } from "./ts/BadTokenMock";
-export { Bytes32LibraryMock } from "./ts/Bytes32LibraryMock";
```
felix2feng: Why remove all of these?

contracts/lib/TimeLockUpgradeV3.sol
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * The TimeLockUpgradeV3 contract contains a modifier for handling minimum time period updates
+ *
+ * CHANGELOG:
+ * - Requires that the caller is the owner
+ * - New function to allow deletion of existing timelocks
+ * - Added upgradeData to UpgradeRegistered event
+ */
+contract TimeLockUpgradeV3 is
+    Ownable
```
felix2feng: Instead of copying all the state and logic, why not inherit `TimeLockUpgradeV2.sol` and then override the timeLockUpgrade function?

contracts/lib/TimeLockUpgradeV3.sol
```diff
@@ -0,0 +1,155 @@
...
+        bytes32 indexed _upgradeHash
+    );
+
+    /* ============ Modifiers ============ */
+
+    modifier timeLockUpgradeV3(
+        address _validator,
+        function(address) internal returns (bool) _check
+    ) {
+        require(
```
felix2feng: This seems a bit over-done. Instead of putting this check here, I'm thinking that V3 would just remove the owner check. And then you just have a separate modifier before the TimeLockUpgrade for the arbitrary check.

artifacts/index.ts
bweick: They aren't used elsewhere, like I thought they would be when I was exporting helpers. So just cleaning up.

contracts/lib/TimeLockUpgradeV3.sol
bweick: Hmmm feels to me like doing that heightens the potential for misusage and allowing someone to call a function that they shouldn't be able to call. Knowing that TimeLockUpgradeV3 doesn't check to make sure the owner is the caller anymore is pretty specific domain knowledge. Whereas if someone has to pass in a verifying function that requires someone to more explicit about what they're doing.

contracts/lib/TimeLockUpgradeV3.sol
```diff
@@ -0,0 +1,90 @@
...
+
+
+/**
+ * @title TimeLockUpgradeV3
+ * @author Set Protocol
+ *
+ * The TimeLockUpgradeV3 contract contains a modifier for handling minimum time period updates
+ *
+ * CHANGELOG:
+ * - Requires that the caller is the owner
```
felix2feng: Update changelog to reflect the actual changes

contracts/lib/TimeLockUpgradeV3.sol
felix2feng: What types of validations are being put here that needs to be generalized?

Another approach is having the check be an unimplemented function that an inherited contract needs to implement

bweick:
@felix2feng can't respond to above comment directly. But I suppose the weakness with that implementation is that we can't do any validations based on the data passed in to the function itself. In this case that means we can't use `_tradingPool` to identify the trader we need to be checking against.

In my mind for any of these more product type contracts where function calls are permissioned but to certain outside parties instead of the "owners" of the contract itself, I can see use cases where you have to take some input (in this case an address) and be able to relate it somehow to the sender or perhaps some other state on the contract.

Ultimately, linking of ownership is generally going to involve addresses since they are "unique". I could also see a version where an identifying hash is maybe passed to the modifier but addresses are essentially just a subset of that and can't think of a place we would really use those at this point.

PR #601, opened 2020-03-05
cToken module audit fixes

richardliang:


contracts/core/modules/CTokenExchangeIssuanceModule.sol
```diff
@@ -450,14 +457,12 @@ contract CTokenExchangeIssuanceModule is
         coreInstance.withdrawModule(
```
bweick: Can we add param names here? Like that suggestion from ABDK

contracts/core/modules/CTokenExchangeIssuanceModule.sol
```diff
@@ -289,13 +290,16 @@ contract CTokenExchangeIssuanceModule is
         uint256[] memory baseSetUnits = baseSet.getUnits();
         uint256 baseSetNaturalUnit = baseSet.naturalUnit();
 
+        // Calculate the number of natural units required and round down to nearest natural unit
```
bweick: Maybe add note that we are guaranteed this is an integer because we check that `quantity.mod(naturalUnit) == 0`

contracts/core/modules/RebalancingSetCTokenExchangeIssuanceModule.sol
```diff
@@ -86,7 +87,7 @@ contract RebalancingSetCTokenExchangeIssuanceModule is
 
     /**
      * Withdraw any base Set components to the user from the contract. If component is a supported cToken,
```
bweick: Unclear to me what purpose of this function is if we aren't unwrapping the cToken. Doesn't that already exist in the inherited contract?

PR #615, opened 2020-03-27
Remove set-protocol-strategies circular dependency

justinkchen:
duplicates some set-protocol-strategies files in order to avoid importing from there especially since it is only for the mock


justinkchen:
the supplementary contracts, in particular are copied from set-protocol-strategies

PR #617, opened 2020-04-01
Flush underlying from cToken exchange issuance

richardliang:


test/contracts/core/modules/rebalancingSetCTokenExchangeIssuanceModule.spec.ts
```diff
@@ -1454,14 +1461,17 @@ contract('RebalancingSetExchangeIssuanceModule', accounts => {
       });
 
       it('returns the user the leftover receive token amount', async () => {
```
bweick: Can we have a test that does the baseSet component and then another one for the underlying component?

contracts/core/modules/RebalancingSetCTokenExchangeIssuanceModule.sol
```diff
@@ -85,6 +85,60 @@ contract RebalancingSetCTokenExchangeIssuanceModule is
...
+                        _returnAddress,
+                        underlyingAddress,
+                        underlyingVaultQuantity
+                    );
+                }
+            }
+
+            // Check if base Set components in vault is greater than 0.
+            if (vaultQuantity > 0) {
+                // Return the unexchanged non cToken components from vault to the caller
```
bweick, line 131: Update comment. This will also return the cToken components right?

PR #622, opened 2020-04-06
Fix to make exports from package smaller

justinkchen:


justinkchen:
potentially breaking change for some repos since you now have to import from the direct location instead of from root. i.e. `const Core = require('set-protocol-contracts/dist/artifacts/ts/Core').Core`

PR #625, opened 2020-04-20
Master TWAP Auction Branch

bweick:


contracts/core/liquidators/impl/AuctionGetters.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    function minimumBid(address _set) external view returns (uint256) {
+        return auction(_set).minimumBid;
+    }
+
+    function remainingCurrentSets(address _set) external view returns (uint256) {
+        return auction(_set).remainingCurrentSets;
+    }
+
+    function startingCurrentSets(address _set) external view returns (uint256) {
```
felix2feng, line 40: It would have been cool to make these prefixed w/ `get` as well. But understand how it wouldn't be backwards compatible

test/contracts/core/liquidators/impl/auctionGetters.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+      return auctionGetters.getCombinedCurrentSetUnits.callAsync(
+        subjectSet,
+        { from: subjectCaller },
+      );
+    }
+
+    it('returns the correct combined current set units', async () => {
+      const combinedCurrentSetUnits = await subject();
+
+      expect(JSON.stringify(combinedCurrentSetUnits)).to.equal(JSON.stringify(auctionState.combinedCurrentSetUnits));
```
felix2feng, line 137: Man, we need a test-helper to do array comparisons that stringifies it for us

PR #626, opened 2020-04-22
Brian/twap impl

bweick:


contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -1,7 +1,8 @@
 /*
     Copyright 2020 Set Labs Inc.
 
-    Licensed under the Apache License, Version 2.0 (the "License");
+    Licensed under the Apache License,
```
felix2feng, line 4: Why 2 lines?

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -29,8 +41,13 @@ import { TwoAssetPriceBoundedLinearAuction } from "../impl/TwoAssetPriceBoundedL
  * TwoAssetPriceBoundedLinearAuction
  */
 contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
+    using SafeMath for uint256;
+    using CommonMath for uint256;
+    using AddressArrayUtils for address[];
+    using BoundsLibrary for BoundsLibrary.Bounds;
 
     /* ============ Structs ============ */
+
```
felix2feng, line 50: Would be good to add javadoc to provide descriptions / definitions of each of the items in the struct

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
         uint256 chunkAuctionPeriod;
         uint256 chunkSize;
     }
+
+    struct TWAPLiquidatorData {
```
felix2feng, line 60: Same here to add docs to describe / define each Struct member

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+        uint256 usdChunkSize;
+        uint256 chunkAuctionPeriod;
+    }
+
+    /* ============ Constants ============ */
+
+    // Auction completion buffer assumes completion potentially 2% after fair value when auction started
+    uint256 constant AUCTION_COMPLETION_BUFFER = 2;
```
felix2feng, line 68: Add explicit `visibility` modifier (e.g. private/public)

felix2feng: Just a thought - This assumption could be void if you have auctions that run away from fair value.

Would this be a variable that we'd ever want to change (e.g. via an admin function)?

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+
+    /* ============ State Variables ============ */
+
+    mapping(bytes32 => BoundsLibrary.Bounds) public chunkSizeWhiteList;
```
felix2feng, line 72: Would define each of these variables w/ a comment.

chunkSizeWhiteList: `Mapping between an address pair's unique keccak256 hash and the min/max USD-chunk size`

expectedChunkAuctionLength: `Estimated length in seconds of a chunk auction`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+    uint256 public expectedChunkAuctionLength;
+
+    /* ============ Constructor ============ */
+
+    /**
+     * TWAPAuction constructor
+     *
+     * @param _oracleWhiteList        OracleWhiteList used by liquidator
+     * @param _auctionPeriod          Length of auction
```
felix2feng, line 81: `Length of auction in seconds`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+     * @param _rangeStart             Percentage below FairValue to begin auction at
+     * @param _rangeEnd               Percentage above FairValue to end auction at
+     * @param _assetPairHashes        Identifying hashes of asset pairs
```
felix2feng, line 84: `List of Keccak256 hashes of asset pairs`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+     * @param _assetPairBounds        Chunk size bounds for each asset pair
```
felix2feng, line 85: `List of USD chunk auction size bounds`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+            _rangeStart,
+            _rangeEnd
+        )
+    {
+        require(
+            _assetPairHashes.length == _assetPairBounds.length,
+            "TWAPAuction: Must be equal amount of asset pair hashes and bounds."
+        );
+
+        for (uint8 i = 0; i < _assetPairHashes.length; i++) {
```
felix2feng, line 108: We conventionally do `uint256` as `uint8` doesn't actually save us anything.

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+            require(
+                _assetPairBounds[i].isValid(),
+                "TWAPAuction.constructor: Passed asset pair bounds are invalid."
+            );
+
+            chunkSizeWhiteList[_assetPairHashes[i]] = _assetPairBounds[i];
+        }
+
+        expectedChunkAuctionLength = _auctionPeriod
```
felix2feng, line 117: Would add a comment to describe what this is doing and why.

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+    function initializeTWAPAuction(
+        TWAPState storage _twapAuction,
+        ISetToken _currentSet,
+        ISetToken _nextSet,
+        uint256 _startingCurrentSetQuantity,
+        TWAPLiquidatorData memory _liquidatorData
+    )
+        internal
+    {
+        // Calculate value of rebalance volume
```
felix2feng, line 142: What is the volume denominated in? Would be useful if you can add it to the comment

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+        // Set TWAPState
+        _twapAuction.orderSize = _startingCurrentSetQuantity;
+        _twapAuction.orderRemaining = _startingCurrentSetQuantity.sub(chunkSize);
+        _twapAuction.chunkSize = chunkSize;
+        _twapAuction.lastChunkAuctionEnd = block.timestamp.sub(_liquidatorData.chunkAuctionPeriod);
+        _twapAuction.chunkAuctionPeriod = _liquidatorData.chunkAuctionPeriod;
+    }
+
+    /**
+     * Resets state for next chunk auction (except minimumBid and token flow arrays), and updates
```
felix2feng, line 174: Good comments overall in this function. They explain the intent of the function

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+        uint256 _chunkAuctionSize
+    )
+        internal
+    {
+        _twapAuction.chunkAuction.auction.startingCurrentSets = _chunkAuctionSize;
+        _twapAuction.chunkAuction.auction.remainingCurrentSets = _chunkAuctionSize;
+        _twapAuction.chunkAuction.auction.startTime = block.timestamp;
+        _twapAuction.chunkAuction.endTime = block.timestamp.add(auctionPeriod);
+
+        // Since currentSet and nextSet param is not used in calculating start and end price on
```
felix2feng, line 216: Good comment here

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+            ISetToken(address(0))
+        );
+        _twapAuction.chunkAuction.endPrice = calculateEndPrice(
+            _twapAuction.chunkAuction.auction,
+            ISetToken(address(0)),
+            ISetToken(address(0))
+        );
+    }
+
+    function validateLiquidatorData(
```
felix2feng, line 230: Add a short javadoc that says what its trying to validate

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+            rebalanceVolume
+        );
+    }
+
+    /**
+     * Validates passed in parameters for TWAP auction
+     *
+     * @param _twapLiquidatorData       Struct of passed liquidator data
+     * @param _assetPairHash            Hash of asset pair being rebalanceds
+     * @param _rebalanceVolume          Value of collateral being rebalanced
```
felix2feng, line 260: Specify USD value

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+    )
+        internal
+        view
+    {
+        require(
+            chunkSizeWhiteList[_assetPairHash].isWithin(_twapLiquidatorData.usdChunkSize),
+            "TWAPAuction.validateTWAPParameters: Passed chunk size must be between bounds."
+        );
+
+        uint256 chunkAuctions = _rebalanceVolume.divCeil(_twapLiquidatorData.usdChunkSize);
```
felix2feng, line 275: Name suggestion: `numAuctions`, `numberAuctions`, or `auctionCount`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+        uint256 expectedTWAPAuctionTime = chunkAuctions.mul(expectedChunkAuctionLength)
```
felix2feng, line 276: Name suggestion: `expectedRebalanceLength`

felix2feng: Would add a comment to describe the intent of the math

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+    }
+
+    /**
+     * Calculates chunkSize of auction in current Set terms
+     *
+     * @param _totalSetAmount       Total amount of Sets being auctioned
+     * @param _rebalanceVolume      The total value being auctioned
+     * @param _usdChunkSize         Value of chunk size
+     */
+    function calculateChunkSize(
```
felix2feng, line 339: I feel like our variables and functions need to do a better just describing its denomination, particularly where things can be denominated in USD, currentSets, etc. It can get confusing if not explicit.

Ideally, we have a typing for this. But if we don't, we can enforce it in the variable name.

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -39,4 +56,357 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
+        // equal to the usdChunkSize
+        if (_rebalanceVolume.div(_usdChunkSize) >= 1) {
+            return _totalSetAmount.mul(_usdChunkSize).div(_rebalanceVolume);
+        } else {
+            return _totalSetAmount;
+        }
+    }
+
+    /**
+     * Parse passed Liquidator data
```
felix2feng, line 358: Update me

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Calculate the rebalance volume as the difference in allocation percentages times market
+     * cap.
+     *
+     * @param _currentSet                       The Set to rebalance from
+     * @param _nextSet                          The Set to rebalance to
+     * @param _oracleWhiteList                  OracleWhiteList used for valuation
+     * @param _currentSetQuantity               Quantity of currentSet to rebalance
+     */
```
felix2feng, line 52: Add return param

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,150 @@
...
+ */
+library LiquidatorUtils {
+    using SafeMath for uint256;
+    using CommonMath for uint256;
+
+    /* ============ Internal Functions ============ */
+
+    /**
+     * Calculate the rebalance volume as the difference in allocation percentages times market
+     * cap.
```
felix2feng, line 46: Would be helpful if you can also put the formula for calculation here?

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,150 @@
...
+        return currentSetValue.mul(_currentSetQuantity).deScale().mul(allocationChange).deScale();
+    }
+
+    /**
+     * Calculate the allocation percentage of passed asset in Set
+     *
+     * @param _setToken             Set being evaluated
+     * @param _oracleWhiteList      OracleWhiteList used for valuation
+     * @param _asset                Asset that's allocation being calculated
+     */
```
felix2feng, line 100: return param

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,150 @@
...
+        // Calculate allocationAsset's current set allocation
+        address allocationAsset = _currentSet.getComponents()[0];
+        uint256 currentSetAllocation = calculateAssetAllocation(
+            _currentSet,
+            _oracleWhiteList,
+            allocationAsset
+        );
+
+        // Calculate allocationAsset's next set allocation
+        uint256 nextSetAllocation = calculateAssetAllocation(
```
felix2feng, line 78: Allocation is denominated in scaled percentage value?

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,150 @@
...
+    {
+        address[] memory components = _setToken.getComponents();
+
+        // Get index of asset and return if asset in Set
+        (
+            uint256 assetIndex,
+            bool isInSet
+        ) = AddressArrayUtils.indexOf(components, _asset);
+
+        // Calculate allocation of asset or return 0 if not in Set
```
felix2feng, line 118: Why can't we just call `calculateTokenAllocationAmountUSD` and divided by `calculateSetTokenDollarValue`?

contracts/core/liquidators/twap-impl/TWAPAuction.sol
bweick: My command key has been screwy haha, must have accidentally done that when I was trying to hot key something

contracts/core/liquidators/twap-impl/TWAPAuction.sol
bweick: I guess in theory it could be anything that's why I didn't specify. Perhaps makes the `usdChunkSize` param a misnomer

contracts/core/liquidators/twap-impl/TWAPAuction.sol
bweick: What would cause us to change it? We foresee more volatility coming up? Feels like something we'd only change in an instant and then want to change back again, but don't think that's what we want to do.

contracts/core/liquidators/twap-impl/TWAPAuction.sol
bweick: Disagree, it should be agnostic because in theory we could use any denomination we want. If anything I think we're actually being overly restrictive and should be NOT explicitly defining it.

contracts/core/liquidators/utils/LiquidatorUtils.sol
bweick: Less efficient

bweick: This way we only have to calculate component value twice instead of three times

contracts/core/liquidators/twap-impl/TWAPAuction.sol
bweick: Why is that name any clearer?

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -29,14 +40,389 @@ import { TwoAssetPriceBoundedLinearAuction } from "../impl/TwoAssetPriceBoundedL
...
+        uint256 failAuctionPeriod = IRebalancingSetTokenV3(msg.sender).failAuctionPeriod();
+
+        require(
+            expectedTWAPAuctionTime < failAuctionPeriod,
+            "TWAPAuction.validateTWAPParameters: Expected auction duration exceeds allowed length."
+        );
+    }
+
+    /**
+     * Validates next chunk auction can begin.
```
felix2feng, line 304: Add a bit more description: `The next chunk auction can begin when the previous auction has completed, there are still currentSets to rebalance, and the auction period has elapsed`

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -29,14 +40,389 @@ import { TwoAssetPriceBoundedLinearAuction } from "../impl/TwoAssetPriceBoundedL
...
+        address[] memory currentSetComponents = _currentSet.getComponents();
+        address[] memory nextSetComponents = _nextSet.getComponents();
+
+        address[] memory assetPairComponents = currentSetComponents.union(nextSetComponents);
+
+        return getAssetPairHash(assetPairComponents[0], assetPairComponents[1]);
+    }
+
+    /**
+     * Parse passed Liquidator data
```
felix2feng, line 397: If the comment doesn't say more than the function name itself, don't really need it

contracts/core/liquidators/utils/LiquidatorUtils.sol
```diff
@@ -0,0 +1,154 @@
...
+        if (isInSet) {
+            uint256 setNaturalUnit = _setToken.naturalUnit();
+            uint256[] memory setUnits = _setToken.getUnits();
+
+            uint256 assetValue;
+            uint256 setValue = 0;
+            for (uint8 i = 0; i < components.length; i++) {
```
felix2feng, line 129: Same as above - `uint256` vs `uint8` for consistency

PR #628, opened 2020-05-02
Add TWAPLiquidator to exports and create interface for querying TWAP …

bweick:
…State params.


contracts/core/interfaces/ITWAPLiquidator.sol
```diff
@@ -0,0 +1,39 @@
...
+pragma solidity 0.5.7;
+
+
+/**
+ * @title ITWAPAuctionGetters
+ * @author Set Protocol
+ *
+ * Interface for retrieving TWAPState struct
+ */
+interface ITWAPAuctionGetters {
```
felix2feng, line 26: Do you need to bump as well?

PR #635, opened 2020-05-13
TWAP Scenario Tests

felix2feng:


test/contracts/core/scenario/twapRebalancingSetScenarios.spec.ts
```diff
@@ -0,0 +1,417 @@
...
+    components: {
+      component1Price: ether(188),
+      component2Price: ether(9000),
+      component1Decimals: 18,
+      component2Decimals: 8,
+    },
+    auction: {
+      chunkSize: ether(1000000),
+      chunkAuctionPeriod: new BigNumber(0),
+      expectedAuctions: 2,
```
bweick, line 138: Think it's worthwhile to change the oracle prices in here too between chunk auctions

test/contracts/core/scenario/twapRebalancingSetScenarios.spec.ts
```diff
@@ -0,0 +1,417 @@
...
+  describe(`${scenarios[0].name}`, async () => {
+    it('should successfully complete', async () => {
+      await runScenario(scenarios[0]);
+
+      const newRebalanceState = await setup.rebalancingSetToken.rebalanceState.callAsync();
+      expect(newRebalanceState).to.be.bignumber.equal(SetUtils.REBALANCING_STATE.DEFAULT);
+    });
+  });
+
+  describe(`${scenarios[1].name}`, async () => {
```
bweick, line 260: What are the most high leverage things we can test here? To me it seems like `rebalanceState`, currentSet`, and most importantly `unitShares`

test/contracts/core/scenario/twapRebalancingSetScenarios.spec.ts
bweick: Think there can be an array of chunkAuction objects that define the price AND what aggregate price we expect it to clear at say 50bps or whatever. Since we have this modeled out we could also create an expectation object that we can check against. That object could have unitShares or whatever other data we want to add.

bweick: If we really want to dig into it we could also include collateral token amounts to check for after each chunkAuction. I believe the model breaks those down as well. That would pretty comprehensively cover the most important part of things, the token flows.

PR #636, opened 2020-05-15
Tweak InitializeTWAPLiquidator to use a orderSize that is a multiple of nat unit

felix2feng:


contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -142,32 +142,63 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
         uint256 rebalanceVolume = LiquidatorUtils.calculateRebalanceVolume(
             _currentSet,
             _nextSet,
             oracleWhiteList,
-            _startingCurrentSetQuantity
+            totalOrderSize
         );
 
         // Calculate the size of each chunk auction in currentSet terms
         uint256 chunkSize = calculateChunkSize(
```
bweick, line 171: Don't think this guarantees that each chunkSize is a multiple of a minimumBid. It guarantees that the totalOrderSize is a multiple of a minimumBid but lets say theres 3 chunks. Then each chunk is gonna have some remainder that's 1/3 (ish) of a minimumBid.

felix2feng: hmm, that's a good point. Think we need to calculate a normalized chunkSize first then back into the total order size

contracts/core/liquidators/twap-impl/TWAPAuction.sol
```diff
@@ -157,18 +166,20 @@ contract TWAPAuction is TwoAssetPriceBoundedLinearAuction {
...
-        LinearAuction.initializeLinearAuction(
-            _twapAuction.chunkAuction,
-            _currentSet,
-            _nextSet,
-            chunkSize
-        );
+        // Normalize the chunkSize and orderSize to ensure all values are a multiple of 
+        // the minimum bid
+        uint256 minBid = _twapAuction.chunkAuction.auction.minimumBid;
+        uint256 normalizedChunkSize = chunkSize.div(minBid).mul(minBid);
```
bweick, line 172: So when we validate the twap parameters we check to make sure that the length of the auction won't exceed the rebalanceAuctionPeriod. Unfortunately that's done before any of this rounding and as a consequence we could under estimate the number of auction iterations required to complete the auction. So we could in theory allow an auction that we might expect to exceed the rebalanceAuctionPeriod.

Obviously that is very much an estimate and it's definitely an extremely narrow edge case but it does exist. Don't know if maybe it makes sense to move that particular validation directly into this function instead of having it validated before initialization takes place.

felix2feng: We could move the check after, I suppose. It doesn't feel like an issue, but in what mathematical issues could this be a problem? We could increase the buffer

PR #637, opened 2020-05-17
Performance Fee Calculator Adjust Fee Fix and Scenario Tests

felix2feng:


contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -258,6 +262,14 @@ contract PerformanceFeeCalculator is IFeeCalculator {
...
             validateProfitFeePercentage(feePercentage);
 
+            // IMPORATNT: In the case that a profit fee is initially 0 and is set to a non-zero number,
+            // the actualizeFee / updateFeeState function does not update the high watermark
+            // Thus, we need to reset the high water mark here so that users do not pay for profit fees
+            // since inception.
+            uint256 rebalancingSetValue = SetUSDValuation.calculateRebalancingSetValue(msg.sender, oracleWhiteList);
+
+            feeState[msg.sender].lastProfitFeeTimestamp = block.timestamp;
+            feeState[msg.sender].highWatermark = rebalancingSetValue;
```
bweick, line 272: This needs to be checked to make sure its higher than the previous high watermark. We don't want to reduce the high watermark by accident

felix2feng: good catch

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -258,6 +262,18 @@ contract PerformanceFeeCalculator is IFeeCalculator {
         } else {
             validateProfitFeePercentage(feePercentage);
 
+            // IMPORATNT: In the case that a profit fee is initially 0 and is set to a non-zero number,
+            // the actualizeFee / updateFeeState function does not update the high watermark
+            // Thus, we need to reset the high water mark here so that users do not pay for profit fees
+            // since inception.
+            uint256 rebalancingSetValue = SetUSDValuation.calculateRebalancingSetValue(msg.sender, oracleWhiteList);
+            uint256 existingHighwatermark = feeState[msg.sender].highWatermark;
```
bweick, line 270: Why put this into its own variable?

felix2feng: Thought it would be more clear

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -37,6 +37,10 @@ import { SetUSDValuation } from "../liquidators/impl/SetUSDValuation.sol";
  * Smart contract that stores and returns fees (represented as scaled decimal values). Fees are
  * determined based on performance of the Set and a streaming fee. Set values can be denominated
  * in any any asset based on oracle white list used in deploy.
+ *
+ * CHANGELOG:
+ * - 5/17/2020: Update adjustFee function to update high watermark to prevent unexpected fee actualizations
+ *              when the profitFee was initially 0
```
bweick, line 43: Maybe add a note on the new require as well

contracts/core/fee-calculators/PerformanceFeeCalculator.sol
```diff
@@ -258,6 +262,23 @@ contract PerformanceFeeCalculator is IFeeCalculator {
         } else {
             validateProfitFeePercentage(feePercentage);
 
+            // IMPORATNT: In the case that a profit fee is initially 0 and is set to a non-zero number,
```
bweick, line 265: IMPORTANT

test/contracts/core/fee-calculators/performanceFeeCalculator.spec.ts
```diff
@@ -566,6 +568,109 @@ contract('PerformanceFeeCalculator', accounts => {
...
+            updatedETHPrice.mul(ether(1)).div(updatedETHPrice).round(0, 3)
+          );
+
+          timeElapsed = customTimeElapsed || ONE_YEAR_IN_SECONDS;
+
+          await blockchain.increaseTimeAsync(timeElapsed);
+          await blockchain.mineBlockAsync();
+        });
+
+        describe('and there is a profit', async () => {
```
bweick, line 603: Think we'd want to test whether the new fee is set here as well

test/contracts/core/fee-calculators/performanceFeeCalculator.spec.ts
```diff
@@ -566,6 +568,109 @@ contract('PerformanceFeeCalculator', accounts => {
         expect(feeState.profitFeePercentage).to.be.bignumber.equal(newFeePercentage);
       });
 
+      describe('when the profit fee is initially 0 and a profit is marked', async () => {
```
bweick, line 571: Don't think we need the and a profit is marked phrase here since it tests that and the complementary case below

PR #647, opened 2020-06-09
cToken bidder changes

richardliang:


contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -36,6 +38,11 @@ import { Rebalance } from "../core/lib/Rebalance.sol";
  *
  * A helper contract that mints a cToken from its underlying or redeems a cToken into 
  * its underlying used for bidding in the RebalanceAuctionModule.
+ * 
+ * CHANGELOG:
+ * - Remove reentrant modifier on bidAndWithdraw. This modifier is already used in RebalanceAuctionModule
```
felix2feng, line 43: Would be good to add month/date/year of change

contracts/helper/RebalancingSetCTokenBidder.sol
```diff
@@ -168,6 +174,44 @@ contract RebalancingSetCTokenBidder is
         );
     }
 
+    /**
+     * Bid on rebalancing a given quantity of sets held by a rebalancing token wrapping or unwrapping
+     * a target cToken involved. The tokens are returned to the user. This function is only compatible with
+     * Rebalancing Set Tokens that use TWAP liquidators
+     *
```
felix2feng, line 181: Would be good to add some language like here to note the exact problem this function is trying to solve.

“During a TWAP chunk auction, there is an adverse scenario where a bidder submits a chunk auction bid with a low gas price and iterateChunkAuction is called before that transaction is mined. When the bidder’s transaction gets mined, it may execute at an unintended price. To combat this, the BidAndWithdrawTWAP function checks that a new chunk auction has not been initiated from the point of bidding. The intended use case is that the bidder would retrieve the Rebalancing SetToken’s lastChunkAuctionEnd variable off-chain and submit it as part of the bid.”

bweick:
LGTM

PR #656, opened 2020-08-07
Fix indentation of numbered list in README

fvictorio:


richardliang (approved):

README.md
```diff
@@ -33,27 +33,31 @@ For legacy users, our audited V1 contracts which can be found [here](https://git
 
 ## Testing
 0. Docker Set up
-Firstly, you need to install Docker. The easiest way is to follow the Instructions on https://docs.docker.com/install/#supported-platforms
 
-You need to pull the docker image that you want to use by using the following command:
+    Firstly, you need to install Docker. The easiest way is to follow the Instructions on https://docs.docker.com/install/#supported-platforms
```
rootulp, line 37:
```suggestion
    Firstly, you need to install Docker. The easiest way is to follow the instructions on https://docs.docker.com/install/#supported-platforms
```

README.md
```diff
@@ -33,27 +33,31 @@ For legacy users, our audited V1 contracts which can be found [here](https://git
...
 
-```
-docker pull ethereum/solc:0.5.7
-```
+    You need to pull the docker image that you want to use by using the following command:
```
rootulp, line 39:
```suggestion
    Pull the docker image that you want to use. For example, run:
```

README.md
```diff
@@ -33,27 +33,31 @@ For legacy users, our audited V1 contracts which can be found [here](https://git
...
 
-If you wish not to set up docker, you can turn off the `docker: true` flag in truffle.js
+    ```
+    docker pull ethereum/solc:0.5.7
```
rootulp, line 42: should we consider upgrading the Solidity version here to something like 0.8.0+? 

^ Disregard, it looks like the contracts in this repo may be pinned to 0.5.7 per https://github.com/SetProtocol/set-protocol-contracts/blob/master/contracts/core/Core.sol#L17

README.md
```diff
@@ -33,27 +33,31 @@ For legacy users, our audited V1 contracts which can be found [here](https://git
...
+    ```
+
+    If you wish not to set up docker, you can turn off the `docker: true` flag in truffle.js
```
rootulp, line 45:
```suggestion
    If you wish to not set up docker, you can turn off the `docker: true` flag in truffle.js [here](https://github.com/SetProtocol/set-protocol-contracts/blob/master/truffle.js#L29)
```

README.md
```diff
@@ -33,27 +33,31 @@ For legacy users, our audited V1 contracts which can be found [here](https://git
 
 ## Testing
 0. Docker Set up
```
rootulp, line 35: [optional][nit]

Can we make this ordered list start with step 1? Step 0 seems a bit confusing

```suggestion
1. Docker Set up
```

rootulp (approved):
Thanks for your contribution fvictorio! I think this PR definitely makes the README more readable as-is. My suggestions are optional.

cgewecke:
@rootulp Suspect fvictorio won't be able to revisit this PR since he is very busy maintaining the tooling at Hardhat atm...

Do you think these improvements could go in as they are?

rootulp:
Yep certainly! I think we can merge this PR as-is and I can follow up with another PR that addresses my nits.

cgewecke:
Admin merging because CI is down here.... These changes only touch the README.

